Issue #50, opened 2013-11-28
"terminal is not fully functional"


bodinsamuel:
Hello,

I'm a relatively new user and I'm pretty much satisfied.
But I keep getting some of this Warning "terminal is not fully functional".

It happen for some git command (as far I know but maybe more) like "git diff".
The behavior is actually not so bugged but I can use my arrow key to navigate, juste space bar.

I wonder if it's a work in progress or a choice by the developpers.

:)

PR #56, opened 2013-11-28
change TERM to xterm

### LINKED ISSUE #44: `screen irssi`, `mutt` - Cannot find terminfo entry for 'msys'.
With all the other amazing things, cmder supports ssh out-of-the-box. However, e.g. `screen irssi` and `mutt` cause the error described in the title.

MrException:
Changing the TERM environment variable to xterm. This will allow you to ssh into another box and run tmux or screen properly.

This should fix #44 and #50.

### DIFF
--- a/vendor/init.bat
+++ b/vendor/init.bat
@@ -28,7 +28,9 @@
 
 :: I do not even know, copypasted from their .bat
 @set PLINK_PROTOCOL=ssh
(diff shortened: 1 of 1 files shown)


Typikall:
when i change term into xterm, then clear comman doesn't more work. Any solution ?

PR #273, opened 2014-09-17
Fixes small Powershell' loader issues

LeoColomb:


thor:
Seems like there's not really anything anyone is waiting for, is there?

LeoColomb:
@thor What do you mean? :smile:

PR #311, opened 2014-11-04
FIX CMDER_ROOT for admin launch

sescandell:
When launching a cmd in admin mode, if `%ConEmuDir%\..\..` contains any space, it breaks the init script.
Thanks to this patch, the path is correctly defined now.

### DIFF
--- a/vendor/init.bat
+++ b/vendor/init.bat
@@ -4,7 +4,7 @@
 
 :: Find root dir
 @if not defined CMDER_ROOT (
-    for /f %%i in ("%ConEmuDir%\..\..") do @set CMDER_ROOT=%%~fi
+    for /f "delims=" %%i in ("%ConEmuDir%\..\..") do @set CMDER_ROOT=%%~fi
 )
 
 :: Change the prompt style


killua99:
Maybe you don't need this comment but. Applying this change works with externals tools like PhpStorm when I setup the console with cmder

FrankPeters:
@killua99 Could you explain how you added Cmder to PhpStorm to use as Terminal? When I use `C:\cmder\Cmder.exe` as the Shell path, the Terminal window opens and then closes but Cmder opens up in a separate window (but not in the correct directory).

killua99:
Hi, I was on holidays I forgot to reply. I set on the terminal "Shell path"

cmd /k "C:\opt\cmder\vendor\init.bat"

Hope it work for you.

FrankPeters:
Hey @killua99 it worked however the prompt is pretty awkward, looks corrupt or something. Well, I'll just continue with a separate instance on my 2nd monitor then, thanks for the reply! :+1:

killua99:
You must to be sure you had this patch in your init.bat file. If not it look broken

FrankPeters:
Oh I see, it's not yet in stable release. Great, then I'll try it again when it's merged into master. Thanks!

zerosurfer:
@killua99 I have the same issue , it looks awkward.
image

PR #326, opened 2014-11-20
changes copied from PR#256

kohenkatz:
Another try to see if the correct branch shows up.

See #256 (and discussion at #312)


bondz:
Is this in master now? Cmder still just shows the previous console on my machine without opening a new tab. Cmder version 1.2 on Windows 8.1 Pro 64 bit.

Issue #191, opened 2014-05-08
Open cmder in 64-bit on 64-bit windows


dfortner:
I am quite a tech-newbie so this is probably due to my own inexperience, but I was wondering how to get cmder.exe to open the 64-bit version of conemu that is included. I have the 64-bit windows and the 32-bit version of cmder is opening.

PR #350, opened 2014-12-17
Add make launcher run ConEmu64 on 64-bit Windows

Shoozza:
This implements #191 for the launcher.
It will run ConEmu64 on 64-bit Windows and ConEmu on 32-bit Windows.

### DIFF
--- a/launcher/src/CmderLauncher.cpp
+++ b/launcher/src/CmderLauncher.cpp
@@ -100,7 +100,18 @@ void StartCmder(std::wstring path, bool is_single_mode)
 
 	PathCombine(icoPath, exeDir, L"icons\\cmder.ico");
 	PathCombine(cfgPath, exeDir, L"config\\ConEmu.xml");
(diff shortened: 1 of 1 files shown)


Shoozza:
Is there a reason we ship ConEmu64.exe then if we never call it from the launcher?

Jackbennett:
@Shoozza I suspect just for simplicity as we just ship the whole conemu download package in /vendor.

LeoColomb:
:+1:

Stanzilla:
Could probably still merge this as it does not hurt and we avoid running a 32bit process if we can run a 64bit one (doesn't matter at all, really)

Stanzilla:
@MartiUK we could still take this, right? See my last comment and might be a good start if we ever want to do split 64bit/32bit zips..

vBm:
It's kinda pointless to run a wrapper when there's a native app imho.

Stanzilla:
https://github.com/cmderdev/cmder/commit/6a474dd6241f78e8df0d0af2b9df18a1ed048281

PR #358, opened 2014-12-24
Rework `alias` command to not to use external tools

vladimir-kotikov:
This includes following changes:
1. Adds handling of duplicate aliases internally by `alias` script instead of external perl script.
2. Adds additional `/d` option to remove already defined alias (and replaces `unalias.bat` -> `unalias` alias :) ).
3. Adds the way to show defined aliases by typing `alias` command without arguments.


vladimir-kotikov:
@MartiUK, @bliker PTAL

bin/alias.bat
```diff
@@ -14,18 +20,28 @@ if not ["%_temp%"] == ["%_temp2%"] (
...
 echo Aliases reloaded
 goto:eof
 
+:p_show
+type "%ALIASES%" || echo No aliases found at "%ALIASES%"
+goto :eof
+
 :p_help
 echo.Usage:
 echo.	alias name=full command
```
narnaud: Maybe replace with:
```
echo.   alias [/reload] [/d] [name=full command]
echo.     /reload  Reload the aliases file
echo.     /d       Delete an alias (must be followed by the alias name)
echo.
echo.   If alias is called with any parameters, it will display the list of existing aliases.
echo.   In the command, you can use the following notations:
```

vladimir-kotikov: Agree. Done in https://github.com/vladimir-kotikov/cmder/commit/dba00b9175c5e90d7a28d562d8dc117f7516c46f

narnaud:
+1 from me, I'm actually using it ;)
THanks

vladimir-kotikov:
You're welcome! :)

MartiUK:
Looks good, thanks @vladimir-kotikov

vladimir-kotikov:
Thanks for merging @MartiUK

PR #368, opened 2015-01-12
Support user/machine-specific config via CMDER_CONFIG env

vladimir-kotikov:
Implementation for https://github.com/bliker/cmder/issues/357


vladimir-kotikov:
Have a question to @MartiUK . Is Cmder executable should be checked into this repo as well as its sources? If so, i will add it to this PR in separate commit.

MartiUK:
Oh god, I forgot. Just the sources please, I need to remove the exe from the repo.

vladimir-kotikov:
I can make this in this PR and add it to .gitignore as well.

bin/alias.bat
```diff
@@ -1,6 +1,7 @@
 @echo off
 
-set ALIASES=%CMDER_ROOT%\config\aliases
+if defined CMDER_ROOT set ALIASES=%CMDER_ROOT%config\aliases
```
narnaud: I think you are missing a \ after %CMDER_ROOT%

launcher/src/CmderLauncher.cpp
```diff
@@ -99,21 +100,29 @@ void StartCmder(std::wstring path, bool is_single_mode)
...
+	SetEnvironmentVariable(L"CMDER_CONFIG", cmderConfig);
+
 	if (is_single_mode) 
 	{
 		swprintf_s(args, L"/single /Icon \"%s\" /Title Cmder /LoadCfgFile \"%s\"", icoPath, cfgPath);
 	}
 	else 
 	{
-		swprintf_s(args, L"/Icon \"%s\" /Title Cmder /LoadCfgFile \"%s\"", icoPath, cfgPath);
+		swprintf_s(args, L"/Icon \"%S\" /Title Cmder /LoadCfgFile \"%S\"", icoPath, cfgPath);
```
narnaud: Just curious, why changing %s into %S?

narnaud:
When reading at first, I thought it was only a way to specify your own ConEmu.xml file.
But actually, you are setting a new config directory. I was more expecting something in the line of setting %CMDER_CONFIG% to a new ConEmu.xml file.

I see a potential issue with this: my use case would be to change the ConEmu.xml file, but keep all the config for clink, as well as the existing aliases **and** whenever Cmder is updated, enjoy update in those files as well.
Note that I have another pull request going into the direction of user settings: https://github.com/bliker/cmder/pull/367

If Cmder goes this way, it would be nice in this case to add a comment to easily initialize it: just a batch file copying the config/* files from cmder, and setting the %CMDER_CONFIG% env variable.

vladimir-kotikov:
> I think you are missing a \ after %CMDER_ROOT%

Dumb mistake. Fixed.

> Just curious, why changing %s into %S?

Have no idea why, but '%s' doesn't worked correctly. It just pasted firs symbol instead of the whole string. Unfortunately, i'm not very experienced in C++, but i'll try to sort out these format issues.
NOTE: some docs [here] (http://msdn.microsoft.com/en-us/library/hf4y5e3w.aspx)

narnaud:
As is, your patch doesn't work or me. Here is the diff to make it works:
```diff
@@ -105,7 +105,7 @@ void StartCmder(std::wstring path, bool is_single_mode)
        DWORD isSet = GetEnvironmentVariable(L"CMDER_CONFIG", cmderConfig, MAX_PATH);
        if (!isSet)
        {
-               PathCombine(cmderConfig, exeDir, L"config\\");
+               PathCombine(cmderConfig, exeDir, L"config");
        }

        PathCombine(cfgPath, cmderConfig, L"ConEmu.xml");
@@ -120,7 +120,7 @@ void StartCmder(std::wstring path, bool is_single_mode)
        }
        else
        {
-               swprintf_s(args, L"/Icon \"%S\" /Title Cmder /LoadCfgFile \"%S\"", icoPath, cfgPath);
+               swprintf_s(args, L"/Icon \"%s\" /Title Cmder /LoadCfgFile \"%s\"", icoPath, cfgPath);
        }

```

The \\ in the CMDER_CONFIG env variable prevent loading clink when it's not set by the user.
The %S doesn't work for me - I have no idea why the %s doesn't work for you, it should.

And after putting more thoughts in this pull request, I think you got a point and I agree with you.
So +1 after fixing the code.

vladimir-kotikov:
@narnaud Fixed. String formatting issue is exists only in VS 2025 preview, so i changes format patterns back to `%s`.

@MartiUK Do you want me to squash all these commits into one or at least reorganize them?

CumpsD:
As per @narnaud comment, would it be possible to also provide a %CMDER_CONFIG% which holds "ConEmu.xml"? In addition to the changes of this PR.

That way you get this PR, and if CMDER_CONFIG is not set, it is "ConEmu.xml" by default. If it is set, that one is used.

That would be non-breaking for everyone who doesn't have it set.
And for people who want to only differentiate ConEmu.xml per machine (for example between high dpi and normal pcs), while keeping their aliasses etc, instead of having to duplicate those directories, they can by simply setting CMDER_CONFIG.

That way we support very flexible scenarios

vladimir-kotikov:
> would it be possible to also provide a %CMDER_CONFIG%

Sorry, @CumpsD but i can't get your point. There is already one `%CMDER_CONFIG%` defined. How would you define another one?

To solve your problem i see two options:

1. Another variable (e.g `%CONEMU_CONFIG%`), but it looks too complicated. And if someone would like to have separate config for clink? Another var (`%CLINK_CONFIG%`)?
2. _Probably the best_ - each tool should accept, that if there is no configuration is found for it in `%CMDER_CONFIG%`, then it need to use configuration from `%CMDER_ROOT%/config`

CumpsD:
Yes, my mistake, I meant CONEMU_CONFIG. You are right that having an env var per tool is not the best way.

Option 2 is indeed the best, and workable, falling back to CMDER_ROOT/config when something is not found

CumpsD:
Looking forward to this to make a new YouTube video for cmder :)

vladimir-kotikov:
Closing this as a stale. Will resend a new one after reworking/rethinking.

PR #372, opened 2015-01-19
Add /PIN switch to launcher

austinwagner:
Switch explicitly sets the AppId to the Cmder.exe path and opens a
window whose icon can be pinned

Fixes bliker/cmder#154


MartiUK:
could you re-pull this into the development branch? thanks.

PR #374, opened 2015-01-21
Updated vendor references

CumpsD:
- Upgrade to git is needed for a security issue in 1.9.4
- Upgrade of ConEmu is needed for High DPI support


MartiUK:
Thanks.

PR #383, opened 2015-02-02
List aliases and show where saved

alexanderchan:
When running alias it is nice to see the current aliases and to see where they are stored in case they need to be updated.  This is similar to what happens when running alias with no parameters from bash.


vladimir-kotikov:
Displaying existing aliases is already implemented in `development` branch. See https://github.com/bliker/cmder/blob/development/bin/alias.bat#L42

alexanderchan:
@vladimir-kotikov awesome, I didn't see the dev branch, looking forward to trying that out.  Thanks for the great work on cmder.

PR #397, opened 2015-02-22
Handle quoted paths

mikesigs:
I was annoyed at having to use short paths all the time in my aliases. It wasn't until I found a command that actually refused to work with a shortpath (running iisexpress.exe) I was determined to find a solution. 

It appears that the reason the quoted paths weren't working was due to the ::validate alias stuff. The for command would stumble on a quoted path, arguing that /foo was unexpected...

The additions on lines 7-8 wrap the input in quotes and then strip the inner quotes. This variable is used in the for loop (instead of %*) for validating the alias. So long as the alias is valid (contains no spaces) then the existing method of appending the new alias to the aliases file works just fine because it still uses %*.

### DIFF
--- a/bin/alias.bat
+++ b/bin/alias.bat
@@ -1,13 +1,16 @@
 @echo off
 if ["%1"] == ["/?"] goto:p_help
 if ["%2"] == [""] echo Insufficient parameters. & goto:p_help
-::validate alias
+
 setlocal
-for /f "delims== tokens=1" %%G in ("%*") do set _temp2=%%G
+::handle quotes within command definition, e.g. quoted long file names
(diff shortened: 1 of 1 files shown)


mikesigs:
The command I used as my test is:
`alias runapi=runas /netonly /user:etoo\Svc_RESTAPITest "c:\Program Files (x86)\IIS Express\iisexpress.exe /site
:EnergyTrust.Services.Ftr /trace:i"`

Stanzilla:
pull requests usually go into the development branch :)

mikesigs:
My bad. I'll resubmit.
________________________________

PR #398, opened 2015-02-23
Handle quoted paths

mikesigs:
I was annoyed at having to use short paths all the time in my aliases. It wasn't until I found a command that actually refused to work with a shortpath (running iisexpress.exe) I was determined to find a solution. 

Quoted paths weren't working whenever the script used "%*". You'd get an error like "File was unexpected at this time." 

The additions on lines 6-7 wrap the input in quotes and then strip the inner quotes. This variable is used in the for loop (instead of %*) for validating the alias. It's also used in the first usage check. 

I also renamed a variable and indented a couple things.

The command I used as my test was:
alias runapi=runas /netonly /user:domain\user "c:\Program Files (x86)\IIS Express\iisexpress.exe /site :WebAppName /trace:i"


mikesigs:
Accidentally sent a pull request yesterday master. Sorry for the noob maneuver. Hopefully this PR is better.

bin/alias.bat
```diff
@@ -1,27 +1,31 @@
...
 :: /d flag for delete existing alias
 if ["%1"] == ["/d"] goto:p_del %*
 if ["%2"] == [""] echo Insufficient parameters. & goto:p_help
-::validate alias
-setlocal
-for /f "delims== tokens=1" %%G in ("%*") do set _temp2=%%G
 
-	set _temp=%_temp2: =%
+:: validate alias
+for /f "delims== tokens=1" %%G in ("%_x%") do set alias=%%G
```
vladimir-kotikov: IMHO, since `%_x%` usage is not needed in arguments check (line 10) then it is better to move its definition to actual usage here.

bin/alias.bat
```diff
@@ -1,27 +1,31 @@
...
 
 set ALIASES=%CMDER_ROOT%\config\aliases
+setlocal
+:: handle quotes within command definition, e.g. quoted long file names
+set _x="%*"
+set _x=%_x:"=%
 
-if ["%*"] == [""] echo Use /? for help & echo. & goto :p_show
+:: check command usage
+if ["%_x%"] == [""] echo Use /? for help & echo. & goto :p_show
```
vladimir-kotikov: Probably no need to replace `%*` with `%_x%` here because if  `["%_x%"] == [""]` then `["%*"] == [""]` will be true as well.

vladimir-kotikov:
:+1:

bin/alias.bat
mikesigs: The reason %_x% is used here instead of %* is because of the surrounding quotes. If you use `if ["%*"] == [""]` and try to run the following alias `alias runapi=runas /netonly /user:domain\user "c:\Program Files (x86)\IIS Express\iisexpress.exe /site :WebAppName /trace:i"` you will receive the error "Files was unexpected at this time", because the quotes surrounding %* get mismatched with the quotes around "c:\Program Files...".

bin/alias.bat
mikesigs: Since the %_x% is needed in the arguments check, the declaration must remain at the top of the file.

Stanzilla:
@mikesigs can you rebase, please?

mikesigs:
@stanzilla. Done.

PR #401, opened 2015-02-23
Adding mercuial prompt

utek:
Similar to git prompt I've added mercurial prompt.

### DIFF
--- a/config/git.lua
+++ b/config/git.lua
@@ -39,6 +39,7 @@ function git_prompt_filter()
         end
 
         clink.prompt.value = string.gsub(clink.prompt.value, "{git}", color.."("..branch..")")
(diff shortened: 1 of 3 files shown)


MartiUK:
:+1: Looks good to me Thanks.

vBm:
Thanks to this PR now we have {hg} visible all the time when using git repos.

I don't understand why did you change config/git.lua ?

vBm:
Seems my comment is gone.
Here it goes again.


Thanks to this PR now we have {hg} visible all the time when using git repos. I don't understand why did you change config/git.lua ?


Quick fix for this would be to completely remove git.lua since it seems that everything was merged into cmder.lua

Stanzilla:
Filed as https://github.com/bliker/cmder/issues/494

PR #405, opened 2015-02-25
ConEmu preview release. Newer clink

Jackbennett:
Lets move to the much newer conemu preview. I've been having many crashes with the current stable so maybe this will work better.

While I'm at it went and added the newer clink too.


Jackbennett:
One day I'll understand what git is going to do. Will remake without the tons of other commits.

PR #406, opened 2015-02-25
Use a -Full parameter to download all sources rather than the minimum

Jackbennett:
Use a full switch to indicate if we want to go and pull a copy of git
since its huge and if a developer is cloning the repo to work on and needs
to build it they already have git.

This isn't the a very robust way to do it but it gets the job done for now. I'm just trying to improve a contributors first experience


MartiUK:
LGTM thanks. :+1:

PR #414, opened 2015-02-28
Added support for gzipped vendor files and some other small improvements.

r3Fuze:
I wanted to add another library to my vendor files, specifically github's [hub utility](https://github.com/github/hub/), but all the downloads are gzipped tar.gz files which need to be extracted twice using 7zip.

The code in utils.ps1 will check if the file is gzipped and then extract it twice using temporary files and other things. The code feels a little hacky and is not very well documented. Would be pretty cool if someone could create some cleaner code or find some improvements for mine.

I also added a `-NoDownload` switch which skips all downloads when building. Very useful for debugging the launcher.

I always had problems with the MSBuild.exe file, which one to use, where to place it and some other things, so I added some code that finds the correct one for you and then uses it instead. `[System.Runtime.InteropServices.RuntimeEnvironment]::GetRuntimeDirectory() + "MSBuild.exe"` Finds the path to the file (example: `C:\Windows\Microsoft.NET\Framework64\v4.0.30319\MSBuild.exe`) and then executes it with `&$msBuild CmderLauncher.vcxproj...`. I'm honestly not sure exactly what the `&` does, but it makes the code work.

And lastly I fixed up the code style a bit. It's still all over the place but I made it a bit more consistent. I wasn't able to find any real style guides.

Let me know if there's anything you don't like or want changed.

### DIFF
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -17,7 +17,11 @@
 
     Executes a full build for cmder, including git. This is equivalent to the "full" style package in the releases
 .EXAMPLE
-    .\build -verbose
+    .\build.ps1 -NoDownload
+
+    Skip all downloads and only build launcher
+.EXAMPLE
+    .\build.ps1 -verbose
 
     Execute the build and see what's going on.
 .EXAMPLE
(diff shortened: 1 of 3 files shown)


scripts/build.ps1
```diff
@@ -47,43 +51,52 @@ Param(
...
-}
 
-Pop-Location
+    Pop-Location
+}
 
 Push-Location -Path $launcher
-msbuild CmderLauncher.vcxproj /p:configuration=Release
+# Using '&' to execute MSBuild.exe
+&$msBuild CmderLauncher.vcxproj /p:configuration=Release
```
Jackbennett: Reading what `&` actually does [here](https://com2kid.wordpress.com/2011/09/25/powershell-call-operator-using-an-array-of-parameters-to-solve-all-your-quoting-problems/) I might suggest to tidy this up givien `&` a list of strings to execute

Jackbennett:
I tried looking by the way and you can't extract tar.gz in one step with 7z. Other than that this works ok, the commit attaches a bit of a mash of things though but they all seem good.

scripts/build.ps1
```diff
@@ -47,43 +51,52 @@ Param(
...
+    # MSBuild.exe location. Acquired using system functions
+    [string]$msBuild = [System.Runtime.InteropServices.RuntimeEnvironment]::GetRuntimeDirectory() + "MSBuild.exe"
 )
 
 . "$PSScriptRoot\utils.ps1"
 $ErrorActionPreference = "Stop"
 
-Push-Location -Path $saveTo
-$sources = Get-Content $sourcesPath | Out-String | Convertfrom-Json
+if (-not $noDownload) {
```
vladimir-kotikov: Double negation makes this check unreadable. Consider invert `if` block.

DRSDavidSoft:
This PR was never merged, so I tried to integrate some features (such as build only, no downloads) and improvements to coding styles in master.

Not integrating the `msbuild` switch since I think `vcvars.bat` should be called first in order to setup the development environment (hence adding MSBuild to PATH). Also, since it doesn't seem that `gzip` files are (as of now) downloaded from vendored sources, I didn't integrate the gzip feature.

Thank you @r3Fuze for your contributions in 2015! 😄

PR #422, opened 2015-03-15
Added git shell task.

ragekit:
Finally I think i found out how to do a correct pull request.

So, added a git shell task, to allo cmder to open directly in a sh

I also corrected identation in config.xml

sole problem : the task don't log commands in .history


MartiUK:
it should log to .bash_history

PR #424, opened 2015-03-16
Allow per-computer ini file.

kodybrown:
Added code to check for the existence of a customized ini file named `ConEmu-%COMPUTERNAME%.ini`, before checking for the standard ini file `ConEmu.ini`.


launcher/src/CmderLauncher.cpp
```diff
@@ -87,6 +87,7 @@ void StartCmder(std::wstring path)
 	wchar_t exeDir[MAX_PATH] = { 0 };
 	wchar_t icoPath[MAX_PATH] = { 0 };
 	wchar_t cfgPath[MAX_PATH] = { 0 };
+	//wchar_t cfgExpPath[MAX_PATH] = { 0 };
```
MartiUK: Could you outright remove this rather than commenting it out, that's why source control is a thing.

MartiUK:
Also, could you redo this pull request into develop?

Thanks.

PR #427, opened 2015-03-17
Added code to check for the existence of a customized ini file..

kodybrown:


MartiUK:
Could you write some documentation about your change, for new users?

kodybrown:
Looks for settings in the following file first `config\\ConEmu-%COMPUTERNAME%.xml`. If that file is not found, then it will look in the default location `config\\ConEmu.xml`.

kodybrown:
I updated the README in my fork, is that what you're talking about? I can add some code-comments also?

kodybrown:
I also added a couple comments in the changed code. Can I update this pull request to use my latest version or do I need to create a new pull request?

Stanzilla:
You can squash and force push to your branch to update the PR

MartiUK:
Just commit and push to your fork, it'll update the PR.

daxgames:
Should this look for ConEmu-%computername%.xml or user-ConEmu.xml?  The latter fits more with what is going on with user config files for the various supported shells.  

I would also suggest an addition to the .gitignore of config/ConEmu-*.xml if we stick with the %computername% route.  user-ConEmu.xml would already be ignored based on the current development .gitignore.

kodybrown:
Do you mean `%USERNAME%-ConEmu.xml`? Using the username wouldn't help me, since I have the same user name on all of my machines. But, I would be fine using `%ComputerName%-ConEmu.xml`... ?

daxgames:
No.  The current development branch supports the following user shell specific config files:

Powershell = user-profile.ps1
Cmder = user-startup.cmd
Bash = user-cmder.sh

I suggested 'user-conemu.xml' to fit the pattern established by the above files and because currently the development '.gitignore' ignores 'config/user-*'. 

I guess I did not understand what you were trying to accomplish, it is clearer now.  It sounds like you are looking for a different conemu config based on the name of the system?  Not sure why but everyone has different needs.

I guess if you wanted to fit the established pattern AND have what you need it could be 'user-%computername%-ConEmu.xml'.

I like configurability/flexibility so I like it what you are doing, I just saw an existing pattern and this did not quite fit the pattern so I thought I'd ask.

kodybrown:
I see what you mean now!

I like being able to pass the settings/config/xml file to the exe as well, but I've found that most things I need are taken care of by using the ComputerName envar. (I have made similar changes to my own versions of other apps too, such as Notepad2.)

I'm totally fine with `config/user-%ComputerName%-ConEmu.xml`. Should I change it before finishing up this PR?

MartiUK:
Is there a need to include the computer name env variable in the filename?

kodybrown:
I think it is still important to have the computer name in the settings file. Although, I think it makes more sense with the computer name at the end.. as in: `config/user-ConEmu-%ComputerName%.xml`.

MartiUK:
Sounds good to me.

PR #436, opened 2015-03-21
Add nodejs and ruby support.

Wind4:
Such changes have a problem?


mickaelandrieu:
Hi @Wind4 , I think this is a little bit out of box of cmder.

But a complete documentation about "how to integrate nodejs and ruby through cmder" can be useful.

PR #440, opened 2015-03-25
git and Posh-git check

Jackbennett:
1. If the minimal client is used there's a chance git doesn't exist on the users machine.
1. Really look for the posh-git module. It seems to need importing because it doesn't have a module manifest. 

I didn't think this looked right so I added some work to it. Somewhere in the build we need to `install-module posh-git` and put it in the `vendor/` path. That will save us adding posh-git to the repo.

See the commit messages for more info I tried to be quite clear.

### DIFF
--- a/vendor/profile.ps1
+++ b/vendor/profile.ps1
@@ -14,13 +14,24 @@ try {
     $gitStatus = $false
 }
 
+function checkGit($Path) {
+    if (Test-Path -Path (Join-Path $Path '.git/') ) {
+        Write-VcsStatus
+        return
+    }
+    $SplitPath = split-path $path
+    if ($SplitPath) {
+        checkGit($SplitPath)
+    }
+}
+
 # Set up a Cmder prompt, adding the git prompt parts inside git repos
 function global:prompt {
     $realLASTEXITCODE = $LASTEXITCODE
     $Host.UI.RawUI.ForegroundColor = "White"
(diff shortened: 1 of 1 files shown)


vendor/profile.ps1
```diff
@@ -5,13 +5,33 @@ if( -not $env:PSModulePath.Contains($CmderModulePath) ){
...
+try {
+    Get-command -Name "git" -ErrorAction Stop >$null
+    Import-Module -Name "posh-git" -ErrorAction Stop >$null
+    $gitStatus = $true
+} catch {
+    Write-Warning "Missing git support"
+    $gitStatus = $false
+}
+
+function checkGit($Path) {
```
LeoColomb: @Jackbennett @MartiUK
I don't understand the reason of this function. Posh-Git checks himself what you write, isn't it?

Jackbennett: I pulled out the code that imported everything all the time from our `vendor\psmodules` in another commit (e811c60ab2433405c0cf12843f56eb2e16425b41)

Posh-git needs a module manifest creating to behave with powershell's auto-loading behaviour. This checks if we have git and need the cmdlet only if we're actually in a git repository. 

All in the name of speed. You don't need to wait for all the git stuff constantly especially if its not even used.

LeoColomb: Oh right. So https://github.com/bliker/cmder/commit/e811c60ab2433405c0cf12843f56eb2e16425b41

PR #460, opened 2015-04-21
echo a new line to the 'alias' config file

airwin:
Echo a new line to the 'alias' config file, after that the new added alias can take effect.

### DIFF
--- a/bin/alias.bat
+++ b/bin/alias.bat
@@ -12,7 +12,7 @@ if not ["%_temp%"] == ["%_temp2%"] (
 	endlocal
 	goto:eof
 )
-
+echo.>> "%CMDER_ROOT%\config\aliases"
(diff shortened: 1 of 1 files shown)


bin/alias.bat
```diff
@@ -12,7 +12,7 @@ if not ["%_temp%"] == ["%_temp2%"] (
 	endlocal
 	goto:eof
 )
-
+echo.>> "%CMDER_ROOT%\config\aliases"
```
DanielGGordon: What is the purpose of the new line? Why is the necessary?

airwin: Without the newline, new alias will be added after the exist one, and then everything is error. :(

MartiUK:
Could you re-pull this into the development branch? Thanks.

PR #492, opened 2015-05-19
Update README.md with link to Visual C++ 2015 Redistr 

OlegGavrilov:
Make sure that people will install Visual C++ 2015 Redistributable libraries before everything.

Btw, doesnt this dependency kill all the "portable" thing?


MartiUK:
@OlegGavrilov I'm looking into it, VC2015 doesn't look very portable, as there's no solid redistributable dlls I can find, I have made the decision to revert back to VC2013 for the time being since the dlls are easily found.

PR #510, opened 2015-05-29
batch file for register right click menu

CodeJuan:
The beginner of windows doesn't know how to run command `.\cmder.exe /REGISTER ALL`, so I write a batch file to run the command automatically  :smiley:


MartiUK:
Could you re-pull this into development? Thanks.

Stanzilla:
Not sure "the beginner of windows" is our target audience.

PR #517, opened 2015-06-04
Remove depreciated Enable-GitColors in posh-git

bondz:
The function Enable-GitColors is depreciated in dahlbyk/posh-git@4e778e2480f584e039b47f0db1c020a197ee9e8f and should be removed here as well. Everytime the console runs, a depreciated warning is shown.


MartiUK:
Thanks.

PR #519, opened 2015-06-05
Check for a custom user shell init script.

glucas:
Allow the user to further customer their shell environment without having to touch files in the vendor directory. 

After the usual setup in init.bat, check if there is a CMDER_USER_INIT variable defined. If found, call the bat file specified.


glucas:
Sorry for the noise -- accidentally created this against the wrong branch.

PR #520, opened 2015-06-05
Check for a user init script.

glucas:
Allow the user to further customize their shell environment without having to touch files in the vendor directory.

After the usual setup in init.bat, check if there is a CMDER_USER_INIT variable defined. If found, call the bat file specified.

### DIFF
--- a/vendor/init.bat
+++ b/vendor/init.bat
@@ -60,4 +60,9 @@
     )
 )
 
+:: Call user-init script, if any
+@if defined CMDER_USER_INIT (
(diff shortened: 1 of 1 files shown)


JanSchulz:
There is now an alternative version in https://github.com/bliker/cmder/pull/612

JanSchulz:
#612 is merged, which executes a user writeable file in the config dir (which does NOT get overwritten on update), not sure if this is still needed

Stanzilla:
If anything else has to be done it has to be on top of what is now in development, I guess that is the best way.

PR #576, opened 2015-08-04
Remove unnecessary .\

avail:
`.\` isn't required on Windows


Jackbennett:
If you're in a PowerShell console you must specify if the executable you want to launch from the current directory is indeed in the current directory with `.\`. in cmd there's no difference.

This is to stop malicious projects just putting a virus in said folder and call it cd.exe to trick the user into running it when they navigate out.

PR #578, opened 2015-08-04
update VS to 2015 release version and switch to /MT for static linking

Stanzilla:
This should make cmder portable again and not require the user to install any VC Redists.


MartiUK:
Thanks.

PR #585, opened 2015-08-09
Revert "Run PowerShell as default"

Stanzilla:
This reverts commit aaf70a4d9fe8b23797e0fdbeb81c1ebe0405e1da.

Changing defaults for users is always kind of a bad idea. In addition, the PowerShell mode wasn't really ready for prime time and causing more trouble than good. Users can still change the default if they want to, but we should ship the best working mode by default and that is cmd.


pavelkouril:
Yes, this is probably a good idea. 

It will solve the "wtf" issues with "`ls` in powershell isn't `ls.exe`" etc., which people who just download cmder and don't know anything about PowerShell will stumble upon.

PR #600, opened 2015-08-18
Use standard path for ConEmu.xml

Maximus5:
Placing xml file in the unusual location restricts or limits users from
* pin cmder to TaskBar;
* using "ConEmu here" feature;
* using "DefTerm" feature.

Details are here: https://github.com/Maximus5/ConEmu/issues/277

### DIFF
--- a/launcher/src/CmderLauncher.cpp
+++ b/launcher/src/CmderLauncher.cpp
@@ -79,6 +79,19 @@ optpair GetOption()
 	return pair;
 }
 
+bool FileExists(const wchar_t * filePath)
+{
+	HANDLE hFile = CreateFile(filePath, GENERIC_READ, FILE_SHARE_READ, NULL, OPEN_EXISTING, 0, NULL);
+
+	if (hFile != INVALID_HANDLE_VALUE)
+	{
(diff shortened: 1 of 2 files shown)


Stanzilla:
Well, looks like you are awesome, thanks :)

r3Fuze:
This becomes a problem when building this project. Running the build scripts will delete the conemu-maximus5 folder before downloading a new version. Also, all the vendor files are ignored from .gitignore so the file will never be checked into git.

I guess this is something that would have to be worked around inside the build script.

Maximus5:
That is not a problem. As you may see, file is copied if it is absent.
And if you want to preserve it on rebuild (?) just add copy back to config before deletion.

Maximus5:
eec3fd5 - Backup and restore ConEmu.xml on `./build.ps1`

Maximus5:
One more drawback of "current" scheme. All settings are **reset** to defatuls when user unpack new cmder version. Proposed patch will preserve user changes made in ConEmu.xml on unpacking cmder.7z into existing installation.

Stanzilla:
Thanks, manually merged it into my fork.

PR #611, opened 2015-08-25
Add a user startup file which can be modified

JanSchulz:
Up to now, init.bat would be overwritten when a user updated
Cmder. The new mechanism will create a new (mostly empty) file
for the user to enter commands which will then be executed on
each following startup.

Closes: https://github.com/bliker/cmder/issues/608


JanSchulz:
Did the PR against the wrong branch :-(

Issue #608, opened 2015-08-25
vendor/init.bat is overwritten with new versions -> use a user startup file?


JanSchulz:
Currently the readme suggests that one should edit /vendor/init.bat to enable ssh-agent and I also used it to manipulate my path. Unfortunately I have to redo this when cmdr is updated to get a upstream updated version of this file. Would it be possible to do something like adding an empty file (with some comments) on first startup which the user can edit and which is executed on each startup? (I do hope that's possible with batch files -> calling a new batch and getting the changes to env variables in the final cmder session)

JanSchulz:
The proposed patch against the development branch is in #612

PR #612, opened 2015-08-25
Add a user startup file which can be modified

JanSchulz:
Up to now, init.bat would be overwritten when a user updated
Cmder. The new mechanism will create a new (mostly empty) file
for the user to enter commands which will then be executed on
each following startup.

Closes: #608

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ All aliases will be saved in `/config/aliases` file
 
 To start SSH agent simply call `start-ssh-agent`, which is in the `vendor/git-for-windows/cmd` folder.
 
-If you want to run SSH agent on startup, uncomment the line in `/vendor/init.bat`so it says `@call "%GIT_INSTALL_ROOT%/cmd/start-ssh-agent.cmd"`.
+If you want to run SSH agent on startup, include the line `@call "%GIT_INSTALL_ROOT%/cmd/start-ssh-agent.cmd"` in `/config/user-startup.bat` (usually just uncomment it).
 
 ## Todo
 
--- a/vendor/init.bat
+++ b/vendor/init.bat
@@ -2,6 +2,9 @@
 :: Sets some nice defaults
 :: Created as part of cmder project
 
+:: !!! THIS FILE IS OVERWRITTEN WHEN CMDER IS UPDATED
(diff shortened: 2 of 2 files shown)


JanSchulz:
This is an alternative version to https://github.com/bliker/cmder/pull/520

JanSchulz:
Ok, this is now involved in a merge conflict. Before trying to resolve the conflict, it would be nice if this solution or the one in #520 is chosen. I can rebase either one :-)

Stanzilla:
I like your approach but I don't think we need the loading messages etc, we don't have them for any other kind of config file that we load.

JanSchulz:
Yo mean the `@echo @echo Executing user startup file: "%%CMDER_ROOT%%\config\user-startup.cmd"` which is written to the user startup file?

I intended that for the user to "find" the new config file (otherwise after the first time, there is no indication that this config file is there) and then let the user remove the line. 

Removing it is no problem...

Stanzilla:
Yes, exactly. Instead of creating the new file inside the script we could just ship it (empty) by default, always call it and add a paragraph to the Readme that tells users how to use it. That's just my idea on how to do it..

JanSchulz:
The problem is that it's overwritten on the next update. At least that's the problem with init.bat and why I added the creation step here.

JanSchulz:
Updated the created script to not print a message to the user. Also updated the readme  (how to enable the ssh agent).

Stanzilla:
Oh indeed, totally forgot about that

JanSchulz:
It would probably be also good to do something similar for the PS init, but I don't know any PS :-(

pavelkouril:
PowerShell wise, this is relevant to "separate user startup file" - https://github.com/cmderdev/cmder/pull/633

JanSchulz:
@pavelkouril Yep, the PS version is more in line of #520: you need to add a env variable to point to you startup file. Not sure if that's more "normal" for PS?

pavelkouril:
It's not an ENV variable - it's a standard PS variable. :)

PS profile is a file that is automatically launched when you launch "classic" PowerShell. However, when using PS from cmder, the PS profile is skipped (and the profile.ps1 is used instead).

PR #621, opened 2015-09-02
posh-hg support.

xavierzwirtz:
Added [posh-hg](https://github.com/JeremySkinner/posh-hg) setup to `profile.ps1`.


xavierzwirtz:
What the hell, looks like I need to change the destination for this merge.

PR #622, opened 2015-09-02
Added posh-hg support for powershell.

xavierzwirtz:
Added [posh-hg](https://github.com/JeremySkinner/posh-hg) setup to `profile.ps1`.


xavierzwirtz:
I am retracing this pr. While it does work, mercurial is so much slower than git that the terminal becomes unusable in mercurial repositories.

PR #633, opened 2015-09-11
If user has his own powershell profile, use it

pavelkouril:
If user has his own `$PROFILE`, cmder should launch it after loading all its settings.

This allows user to set default directory/create new functions or aliases/etc. quite easily and have them available in both cmder and traditional powershell window.

### DIFF
--- a/vendor/profile.ps1
+++ b/vendor/profile.ps1
@@ -50,3 +50,7 @@ if (Test-Path Env:\CMDER_START) {
 } elseif ($Env:CMDER_ROOT -and $Env:CMDER_ROOT.StartsWith($pwd)) {
     Set-Location -Path $Env:USERPROFILE
 }
+
+if (Test-Path $PROFILE) {
+    . $PROFILE
+}


JanSchulz:
Is `$PROFILE" a standard way to extend a powershell console environment?

pavelkouril:
http://www.howtogeek.com/50236/customizing-your-powershell-profile/
http://www.computerperformance.co.uk/powershell/powershell_profile_ps1.htm
https://technet.microsoft.com/en-us/magazine/2008.10.windowspowershell.aspx

Definitely looks like it.

MartiUK:
Could you re-pull this into development please.

Jackbennett:
There's actually 4 profiles to be aware of as my personal extensions reside within currentProfileAllHosts

`$profile | fl -Force`

Hosts being what program is playing host to the powershell session. E.g. 

Run `powershell.exe -noExit -command "Get-host"`
Run `Get-host` from within the powershell ISE you'll get a different name. And can have a different profile.

So this PR just needs checking against what files are called in which order depending on what the user has present.

pavelkouril:
@MartiUK Hmm, so should I open different PR with another branch?

@Jackbennett So, what exactly should be changed? I'm sorry but my understanding of PS is kinda limited (I use it for basic scripting, mostly). If you either could explain what should be done differently (or prepare a more correct PR), it would be wonderful. :)

MartiUK:
@pavelkouril Yes, please.

PR #654, opened 2015-10-06
Update README.md

andrerpena:
Making the download instruction a link to the releases

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ The main advantage of Cmder is portability. It is designed to be totally self-co
 
 ## Installation
 
-1. Download the latest release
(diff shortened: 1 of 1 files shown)


Jackbennett:
Maybe `https://github.com/cmderdev/cmder/releases/latest` would be a better URL. I'll leave that call up to a more active maintainer I'm not sure how @MartiUK wants to tag releases. The powershell in 1.2 didn't sound like it went as smooth as I should have helped make it.

Stanzilla:
Did it a bit different but it points to /releases/ now.

andrerpena:
Thanks

PR #665, opened 2015-10-14
Build from behind proxy & appveyor

### LINKED ISSUE #79: Build script does not work through proxy
Net::HTTP does not automatically pick up proxy settings. The build script should at least allow setting the proxy (by using the http_proxy environment variable?). 

It would be helpful if it automatically retrieved the system proxy settings too (requires reading a WPAD file if system is set to autoconfig).

MartiUK:
Fixes #79.


MartiUK:
The big FAIL means it's worked YAY :tada:

MartiUK:
Really squashed that one.

Issue #505, opened 2015-05-25
powershell read user profile


harkx:
Using 150513 it seems cmder does not use the Powershell profile.
I have a function in that profile to connect to Ofice365, it would be great if launching cmder reads the powershell proflile.

In my case it's located in:

```
C:\Users\me\Documents\WindowsPowerShell\Microsoft.PowerShell_profile.ps1
```

You can query this location in Powershell with this command:

```
$Profile
```

strig:
Go to `settings > startup > tasks` and remove the `--NoProfile` from the PowerShell command arguments.

harkx:
Perfect! Thanks!

nevion:
@strig why is this not default?

Stanzilla:
I guess one could argue that to keep cmder properly portable it should ignore the local PS profile, not sure though

JanSchulz:
Would it be a nice compromise to add someting like this into `user_profile.ps`:

```
# cmder runs the powershell with '--NoProfile', so $Profile isn't called on startup.
# If you want to use your $Profile, you can either uncomment the next lines or 
# change the startup to use $Profile (in cmder under settings > startup > tasks -> 
# the --NoProfile from the PowerShell command arguments).
#if (Test-Path $PROFILE) {
#    . $PROFILE
#} 
```

Jackbennett:
What about a separate pre-configured task that doesn't have `--noProfile`. People can test with it and just move the default over if they choose.

AlexHankins:
I was very puzzled why running PowerShell from the Start menu behaved differently from running it in Cmder.  This cost me some time and effort to research the issue.  Please fix.  Thank you.

daxgames:
Add to the `%cmder_root%\user_profile.ps1`:

```
. $profile
```

PR #666, opened 2015-10-14
Add user startup file for powershell

Pireax:
The cmd version has a customizable file but powershell was missing one.

### DIFF
--- a/vendor/profile.ps1
+++ b/vendor/profile.ps1
@@ -63,3 +63,13 @@ if (Test-Path Env:\CMDER_START) {
 
 # Enhance Path
 $env:Path = "$Env:CMDER_ROOT\bin;$env:Path;$Env:CMDER_ROOT"
+
+
+$CmderUserProfilePath = Join-Path $env:CMDER_ROOT "config/user-profile.ps1"
+if(Test-Path $CmderUserProfilePath) {
+    # Create this file and place your own command in there.
+    Invoke-Expression $CmderUserProfilePath
+} else {
+    Write-Host "Creating user startup file: $CmderUserProfilePath"
+    "# Use this file to run your own startup commands" | Out-File $CmderUserProfilePath
+}


JanSchulz:
`vendor\init.bat` got a comment not to change it at the top of the file. Maybe `vendor/profile.ps1` should also get such a comment?

Stanzilla:
@JanSchulz agree, done.

pavelkouril:
Why not use the user profile though? Maybe @Jackbennett could say which option would be better?

Stanzilla:
Related to https://github.com/cmderdev/cmder/issues/505#issuecomment-148135474 then.

Pireax:
It isn't a question of if it's better to use a user profile or a portable profile file, with the portable profile file the option to use the user profile remains. This way we have both. The user can choose to run the user profile from the portable script or omit the -NoProfile launch option.

Stanzilla:
Yup

Pireax:
Mine too ;)

Jackbennett:
I think using a cmder profile would be good. Anyone who wants to pull in their own PS Profile should know they can just dot source it `. $profile.currentuserAllHosts`

Although if you don't know powershell you won't have your own customized profile anyway.. hmm. Should cmder be it's own self contained little terminal or be a great baseline to add your own stuff onto that also happens to run portably. 

Tough that it's not really I code problem. I vote use a cmder profile but keep in mind it's not a mandatory one. That's consistent with how cmd behaves and also doesn't have to worry about clobbering someone's main profile. At least until someone complains with a better/more informed use case.

pavelkouril:
Yeah, true, you just can launch the $profile from this user startup file. Don't know why this didn't occur to me. :)

AlexHankins:
I don't have to do any special configuration for Git bash; why do I have to do this for Powershell?

PR #667, opened 2015-10-15
Fixed problem with Invoke-Expression

Pireax:
Invoke-Expression has a problem with spaces in path names(oops), this fixes that issue.
Also, the file or atleast the lambda needs to be encoded in UTF-8-BOM else it won't show properly.

### DIFF
--- a/vendor/profile.ps1
+++ b/vendor/profile.ps1
@@ -1,4 +1,4 @@
-# Init Script for PowerShell
+﻿# Init Script for PowerShell
 # Created as part of cmder project
 
 # !!! THIS FILE IS OVERWRITTEN WHEN CMDER IS UPDATED
--- a/vendor/profile.ps1
+++ b/vendor/profile.ps1
@@ -74,7 +74,7 @@ $env:Path = "$Env:CMDER_ROOT\bin;$env:Path;$Env:CMDER_ROOT"
 $CmderUserProfilePath = Join-Path $env:CMDER_ROOT "config/user-profile.ps1"
 if(Test-Path $CmderUserProfilePath) {
     # Create this file and place your own command in there.
-    Invoke-Expression $CmderUserProfilePath
+    . "$CmderUserProfilePath"
 } else {
     Write-Host "Creating user startup file: $CmderUserProfilePath"
     "# Use this file to run your own startup commands" | Out-File $CmderUserProfilePath


vendor/profile.ps1
```diff
@@ -74,7 +74,7 @@ $env:Path = "$Env:CMDER_ROOT\bin;$env:Path;$Env:CMDER_ROOT"
 $CmderUserProfilePath = Join-Path $env:CMDER_ROOT "config/user-profile.ps1"
 if(Test-Path $CmderUserProfilePath) {
     # Create this file and place your own command in there.
-    Invoke-Expression $CmderUserProfilePath
+    & $CmderUserProfilePath
```
Jackbennett: If that's to just run the profile.ps1 into the current session you want to dot source it:

`. "CmderUserProfilePath"`

Honestly I'm not sure what `&` will do different but you'll see a lot of 'dot source' when looking at powershell examples so it would be more consistent.

Pireax:
I looked it up, the difference between dot sourcing and the ampersand is that dot sourcing gives the command prompt access to the variables defined in the script after execution, with an ampersand you can't access them from the command prompt.

vendor/profile.ps1
Pireax: I looked it up, the difference between dot sourcing and the ampersand is that dot sourcing gives the command prompt access to the variables defined in the script after execution, with an ampersand you can't access them from the command prompt.

Stanzilla:
AH so that was what @Jackbennett removed here https://github.com/cmderdev/cmder/commit/7ebfffc9da6c057485f79ec65d35226b14be0c05 :D

Jackbennett:
I wasn't sure what it was but powershell wouldn't run with it so out it came :)

How'd that get in there I thought git would deal with all that. Are you writing PS in something that isn't the ISE? That complained about the character immediately. (Not using the ISE is perfectly fine by the way)

MartiUK:
A Microsoft made application, that doesn't support UTF-8? Colour me surprised. :laughing:

Jackbennett:
I have `out` aliased to `out-file -encoding UTF8` haha. Don't start me on excel and csv's.

Stanzilla:
I actually use GitHub's Atom editor for almost anything

Jackbennett:
Ever given brackets ago @Stanzilla ?

I only ask because I made the jump from sublime to brackets. Atom came out about a year later and I just haven't considered switching.

Only downside with not using the ISE is you're missing that sweet sweet intellisense

Stanzilla:
I always wanted to but was too lazy to switch from Sublime, now Atom has most of the features I wanted to try Brackets for as plugins...Atom is still pretty new but community and devs are great and pushing really hard to get stuff done and I love that, so I'm stuck there for now :D

MartiUK:
I did those last few commits to martiuk-dev with vim... on my phone. Is that too far?

I usually use vim on Linux or sublime text 3 on Qindows.

DanielGGordon:
Vim on your phone? Lol that's hardcore dude

Issue #577, opened 2015-08-04
profile.ps1 doesn't have CMDER_START when duplicating tab in windows 10


drulia:
I have installed windows and cmder, I haven't used this version on a previous windows setup, so cannot say if this is purely windows 10 problem.
Edit: previously I have only used cmd, as powershell wasn't supported the way it is now.

To the problem, when I duplicate current powershell tab, it opens in the default user location. Same happens if I execute `cmder /single /`. It does however work when I execute `cmder /start /`, then it open in the location I've passed.

I've added the line `Write-Host $Env:CMDER_START` in profile.ps1 file  just before it checks it and I see that when duplicating tab or creating in the same window a new tab, it doesn't have a `CMDER_START` and defaults to user profile location.

kermage:
Try this. Goto (**`Settings` > `Startup` > `Tasks`**).
Remove the *`-new_console`* parameter in default predefined task **`{Powershell}`**.

```
PowerShell -NoLogo -NoProfile -NoExit -Command "Invoke-Expression '. ''%ConEmuDir%\..\profile.ps1'''"
```

drulia:
This has one change that I can see, when executing command `cmder /single /`, it opens the new tab in the current directory rather than user profile, but it still ignores the directory I've passed. It is however a slightly more convenient behavior. It doesn't make any difference when duplicating a tab, tab opens in user profile. 

Thanks for trying to help.

kermage:
If found this is not getting through...
https://github.com/bliker/cmder/blob/master/launcher/src/CmderLauncher.cpp#L115

DoCode:
but CMDER_START is not set... why?

MartiUK:
#772

PR #673, opened 2015-10-20
Enable the '/single' switch (#577)

DoCode:
Fixed issue #577


vBm:
@DoCode since .cpp is using tabs and your patch is using spaces, maybe you should fix that and force-push that again (just to be consistent with rest of the code) ? :)

DoCode:
if it is absolutely necessary;)

MartiUK:
@DoCode CMDER_START is still not being set as of the latest development build.

DoCode:
@MartiUK CMDER_START is set to Home when NOT in single mode
https://github.com/cmderdev/cmder/blob/development/vendor/profile.ps1#L105

PR #698, opened 2015-11-08
Fix batch files

daxgames:
Replaced %CMDER_ROOT%\ with %CMDER_ROOT% because the variable already ends with \

### DIFF
--- a/Cmder.bat
+++ b/Cmder.bat
@@ -1,3 +1,4 @@
 @echo off
 SET CMDER_ROOT=%~dp0
+if "%CMDER_ROOT:~-1%" == "\" SET CMDER_ROOT=%CMDER_ROOT:~0,-1%
 start %~dp0/vendor/conemu-maximus5/ConEmu.exe /Icon "%CMDER_ROOT%\icons\cmder.ico" /Title Cmder /LoadCfgFile "%CMDER_ROOT%\config\ConEmu.xml"
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ Thumbs.db
 build/
 Version v*
 *.bak
+config/user-*
--- a/vendor/init.bat
+++ b/vendor/init.bat
@@ -9,6 +9,8 @@
     for /f "delims=" %%i in ("%ConEmuDir%\..\..") do @set CMDER_ROOT=%%~fi
 )
 
+@if "%CMDER_ROOT:~-1%" == "\" SET CMDER_ROOT=%CMDER_ROOT:~0,-1%
+
 :: Change the prompt style
 :: Mmm tasty lamb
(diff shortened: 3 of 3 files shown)


Cmder.bat
```diff
@@ -1,3 +1,3 @@
 @echo off
 SET CMDER_ROOT=%~dp0
-start %~dp0/vendor/conemu-maximus5/ConEmu.exe /Icon "%CMDER_ROOT%\icons\cmder.ico" /Title Cmder /LoadCfgFile "%CMDER_ROOT%\config\ConEmu.xml"
+start %~dp0/vendor/conemu-maximus5/ConEmu.exe /Icon "%CMDER_ROOT%icons\cmder.ico" /Title Cmder /LoadCfgFile "%CMDER_ROOT%config\ConEmu.xml"
```
JanSchulz: This looks like an oversight: the normal launcher does not use /LoadCfgFile anymore

daxgames:
I am also working on enabling the use of the new  git bash.  Should be ready soon

daxgames:
It will be a separate pull request though

Stanzilla:
I don't even know if we want to keep the bat around, it's only used for Windows XP anyway, no?

daxgames:
It is in the source so i fixed it.  Do with it whatever you feel is necessary.

Stanzilla:
I think it looks weird without the `\`in the paths :/

daxgames:
It might look wierd but it is correct.

It looks even wierder when you look at the %path% or a variable that has 
'\\' in it.

Stanzilla:
Well actually it might be the better way to just remove the `\` from the variable, don't you think?

daxgames:
We could do that to get rid of the wierdness.

That would also require changes to my other pull request as it behaves in a manner consistent with what init.bat does and expects %CMDER_ROOT% and $CMDER_ROOT both end with either a '\' or '/'.

Stanzilla:
@cmderdev/trusted-contributors @cmderdev/owners any objections to changing it like that?

Also @daxgames can we leave https://github.com/daxgames/cmder/commit/e56e00669113b9bad3d57b19a3623ebd445191b2 out of this PR, I don
t really think we need your personal files in the .gitignore, I know it does not hurt either but yeah.

daxgames:
I actually modified the .gitignore in my last commit so it is relevant to all and does not contain any personal files.  I only added the config/user-* files that are auto created by launching shells in this build.

Stanzilla:
Nope, that's good.

daxgames:
Squashed

daxgames:
hang on something got messed up in the squash let me fix it.

daxgames:
Squash fixed.  One commit 3 files changed.  Sorry about that somehow I combined two branches. in the previous squash

Stanzilla:
Great, thank you!

PR #701, opened 2015-11-09
Added support for msys2 bash in the new git for windows

daxgames:
Added support for MSys2 Bash.
Renamed cmd task to cmder.
Also added personal files to .gitignore so they never get uploaded to the repo.


daxgames:
Making one more change.

PR #718, opened 2015-11-18
Publish appveyor artefacts

### LINKED ISSUE #717: Change appveyor.yml to publish all resulting artifacts from builds.
This will be useful for new releases and testing pull requests.

MartiUK:
Fixes #717


vBm:
Would've been better if you bothered enough to actually squash those useless million commits into one that actually adds the desired feature instead of just polluting history with junk.
After all, that's the reason you have your own fork and you have PR's xD

MartiUK:
@vBm It won't happen again!

PR #721, opened 2015-11-19
Updating README's latest release to 1.2.9

alexbeardsley:
Updating README's latest release to 1.2.9 because 1.2 often did not work out of the box due to use of Powershell.


Stanzilla:
We already have that in `development`

PR #723, opened 2015-11-20
Release 1.3

MartiUK:
Please contribute to the changelog here if possible @cmderdev/owners @cmderdev/trusted-contributors 
Test archives here (When finished) : https://ci.appveyor.com/project/MartiUK/cmder/build/1.0.90-master/artifacts


Stanzilla:
@MartiUK @cmderdev/trusted-contributors  what do we want to be on the release notes? We should probably wait with releasing until there is a new clink release but maybe we can start drafting the notes already so it does not pile up.

I'd suggest a section just for the vendor stuff that was bumped to the latest version but do we want to list *their* changelogs, too? Or just mention which Cmder issues have been fixed by including a newer version? Kinda annoying to track down.

daxgames:
I know I am not a trusted-contributer, maybe someday, but I am a big fan and daily user of Cmder.

I think we should mention this is probably the most complete Cmder release to date supporting Cmder enhancements/portability in all three shell types, but that is probably stating the obvious.

PR #731, opened 2015-11-24
Fix icons

daxgames:
Added Git for Windows icon to mintty tasks to differentiate them from 100% supported Cmder tasks

Get 100% supported Cmder task icons from icons\Cmder.ico so tabs have an ico if launched from the cmder.bat file

### DIFF
--- a/config/ConEmu.xml
+++ b/config/ConEmu.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <key name="Software">
 	<key name="ConEmu">
-		<key name=".Vanilla" modified="2015-11-09 17:04:58" build="151025">
+		<key name=".Vanilla" modified="2015-11-24 14:43:35" build="151119">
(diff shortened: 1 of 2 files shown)


Stanzilla:
Was the current way of trying to get the icon from the exe not working?

daxgames:
It would probably work fine if I could build the cmder.exe but I can't because I do not have the tools.  I and anyone else trying/developing on the dev branch that cannot build the exe use the cmder.bat to start cmder.

I just did what I did so it would work all the time with or without the cmder.exe.  To me working all the time was better than working some of the time even if the cmder.bat is deprecated.

Stanzilla:
Makes sense

MartiUK:
Any possibility of moving that last commit to a separate pull request? It has nothing to do with fixing the icons :smile:

daxgames:
Sure.  Will do later today.

daxgames:
Done

MartiUK:
Thanks.

PR #734, opened 2015-11-24
Added check for git install path in init.bat.

chase-miller:
Added check for %USERPROFILE%\AppData\Local\Programs\Git path in init.bat.  Installing from https://git-scm.com/download/win seems to put git in this directory.

Long story short, I was having issues doing npm installs pointing to cmder's packaged version of git-for-windows, but the version from https://git-scm.com/download/win worked just fine.  This change defaults to using an installed version of git outside of cmder's packaged version.

### DIFF
--- a/vendor/init.bat
+++ b/vendor/init.bat
@@ -37,6 +37,8 @@
     set "GIT_INSTALL_ROOT=%ProgramFiles%\Git"
 ) else if exist "%ProgramFiles(x86)%\Git" (
     set "GIT_INSTALL_ROOT=%ProgramFiles(x86)%\Git"
+) else if exist "%USERPROFILE%\AppData\Local\Programs\Git" (
+    set "GIT_INSTALL_ROOT=%USERPROFILE%\AppData\Local\Programs\Git"
 ) else if exist "%CMDER_ROOT%\vendor\git-for-windows" (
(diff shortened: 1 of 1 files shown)


MartiUK:
Looks good to me, thanks!

chase-miller:
It turns out that the root cause of the issue I had deals with a bug in ConEmuHk DLL truncating stdout to 4096 characters...not sure how my change fixed it at all (it was only temporary).  See https://github.com/bower/bower/issues/852#issuecomment-141724402.  

I'm going to check out the ConEmu project to see if they already have a fix.  That said, how do we go about adding ConEmu fixes into cmder?

chase-miller:
As an FYI, I copied the latest stable versions of ConEmuHk.dll and ConEmuHk64.dll into the appropriate folder within cmder\vendor and it managed to resolve the issue.  I've got no idea if it's going to create other issues, though.  

Should I open an issue regarding this?

MartiUK:
We try to keep up to date, especially with ConEmu as it has a very rapid release cycle. I can see we're a version behind already in the development branch.

I've had an idea for a dependency updater, but I haven't started any work on it.

chase-miller:
I've created https://github.com/cmderdev/cmder/issues/739 to track this issue :-)

Issue #694, opened 2015-11-05
files in config get overwritten


janschulz:
~~I my case it's problematic for `aliases` but the other files might also be problematic. IMO `config` should be empty apart form a readme.md file and either be populated by `vendor\init.bat` or `cmder` itself.~~

Only the conemu.xml file is overwritten, everything else is generated on first start.

JanSchulz:
I think at least two things should be moved out of the config dir: `aliases` and `cmder.lua`.

For the latter it's I think the easiest to move it to `vendor/cmder.lua` and add a line to load the file which loads `config/config.lua` if it exist. 

Not sure what to do about the `aliases` file, if it is `cmd` only, it could be copied in `init.bat`

daxgames:
Is it possible for the config/aliases file to stay where it is so the cmder team can include 'default' aliases and also the use of a user-aliases file in the config folder that is loaded after the config/aliases if it exists not overwritten on update.  

The alias.bat file would need to be modified to add user aliases to config/user-aliases instead of config/aliases.

Not even sure if it is possible, I'm not a doskey guru, but this provides the best of both worlds.

JanSchulz:
I currently use `config\aliases_jasc` and source that in my `user_startup.bat` file. AFAI understand it, the current aliases are examples, not things the user should have/ are needed by cmder. 

IMO: move the current `config\aliases` file to `vendor\aliases.example` and let `vendor\init.bat` copy that to the config dir, if it does not exist. And then let doskey load it.

JanSchulz:
If #736 gets merged, only `ConEmu.xml` and `settings` are in there. Not sure how to handle them, if it should go the same way as #735/#736, then cmder.exe must do the copy. 

But on the other hand, I've actually no idea how the `ConEmu.xml` file is handled (https://github.com/cmderdev/cmder/issues/695). And actually, after the changes in the new release, I don't think I have to do anything in this files, so I'm fine after these two PRs :-)

daxgames:
What do you guys think of my suggestion of having a cmder team controlled aliases file that is loaded first then a config\user-aliases file that is loaded second?  This would require making the bin\alias.bat manage config\user-aliases.

I have verified this would work and it is an easy change.

I have about a lot of aliases, mostly Git related, that make most Git commands 1-5 characters for both cmd and bash.  I could also look at making these work Powershell.  It would be nice to add these to Cmder.

[...]

PR #735, opened 2015-11-24
Do not overwrite aliases on update

JanSchulz:
This partially adresses https://github.com/cmderdev/cmder/issues/694


JanSchulz:
damn, that was too fast... new PR incoming with the rest of the changes...

PR #740, opened 2015-11-26
Cmder exinit

daxgames:
From the vendor/cmder_exinit file:

Copy this file to your externally installed meaning not in the cmder/vendor
folder Cygwin/Msys2/Msys installs '/etc/profile.d/' folder to integrate
the externally installed Unix like environment into Cmder so it has
access to settings stored in Cmder/config folder when launched.

The destination file extension depends on the shell you use.  For example:

bash - Copy to /etc/profile.d/cmder_exinit.sh
zsh  - Copy to /etc/profile.d/cmder_exinit.zsh


.gitignore
```diff
@@ -13,3 +13,4 @@ build/
 Version v*
 *.bak
 config/user-*
+config/aliases
```
JanSchulz: IMO this could be changed to `config/*`

daxgames:
I made a suggestion in #694.

daxgames:
Wait on this one I need to check some stuff before merging.

daxgames:
This is good to go now.  I was seeing some init funkyness and determined it was unrelated to this change.  This has been tested with Babun(Cygwin), msys, msys2, and the Git for Windows SDK(msys2)

daxgames:
@MartiUK - Should I close this?  Is there any chance it will make it in?  If it might did you see my profile.d PR #855.  If this might make it in I would want to add the profile.d bits to it as well.

MartiUK:
@daxgames go for it

daxgames:
Updating with profile.d like support in #855 - Do not merge until I release here.

daxgames:
This is ready whenever you are.

PR #745, opened 2015-11-28
Do a more thorough check for SSH keys

QWp6t:
This looks for all *.pub keys and then grabs associated private key, ezpz.

### DIFF
--- a/bin/agent.cmd
+++ b/bin/agent.cmd
@@ -28,7 +28,7 @@ SET SSH_AUTH_PID=%VAR%
  
 REM Check if ssh keys are known
 SET KEYS=
-FOR /f "tokens=*" %%I IN ('DIR /B "%HOME%\.ssh\*_rsa"') DO CALL:CHECKKEY %%I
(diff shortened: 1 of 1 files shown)


JanSchulz:
The latest dev version (branch "development") uses/advertises the agent script from git directly. Please check if that works and if not it's probably best to report it there.

QWp6t:
I just tested with that. It misses certain keys. I store keys based on email address.

So `janschulz@example.com.pub` and `janschulz@example.com` would be a public and private key, respectively. The implementation by git-for-windows would not automatically add that key. My implementation in this PR, however, does add that key.

If you do not foresee others having this issue, I can simply add the code to my `user-profile.cmd` file, np. Otherwise if you are interested in discussing this issue, let me know, and I'll make sure to target the correct branch next time.

JanSchulz:
If you want ot have upstream support for that, it's best to add this to the git-for-windows repo (aka here: https://github.com/git-for-windows/MINGW-packages/blob/master/mingw-w64-git/start-ssh-agent.cmd). The agent script is gone in cmder, so cmder cannot do anything about it...

If your setup is so uncommon (I've no idea, I just followed an online tutorial and it worked...) that it is easier to do it on your side with a line of code, then just do it in `user-profile.cmd`. :-)

PR #746, opened 2015-11-28
README.md Updates

daxgames:


DanielGGordon:
Very nice. i see you put in info about mintty. That is very helpful (or at least it was to me). And the example of adding solarized - awesome.

Issue #793, opened 2016-01-09
Update clink settings


JanSchulz:
It seems that there are a lot more settings for clink than the oneswhich are in config/settings: https://github.com/mridgers/clink/blob/be922add219458027285762708f2642eaadb21e3/docs/clink.md#settings

I was finding this as I wanted to find out why clink does not save the historoy after each command but only when the commandline is closed (and so losing history if cmd is killed). Turns out that there is a setting for that: `history_io`...

JanSchulz:
In addition, this should also get a vendor default which is copied to the config dir if it doesn't not exist there... Not sure how to do that, the only place to do that would be the cmder startup itself or duplicate the check+copy in all startups (bash+powershell+cmd)

JanSchulz:
Ok, there is a problem here: clink since 4bfdf04d59dab80e1ce76d7125c546006273e188 is using `vendor` as profile dir, so `.history` and `setting` are created there and used form there :-(

JanSchulz:
Closed in #794

PR #794, opened 2016-01-09
Reorganize how clink settings are loaded

JanSchulz:
4bfdf04d59dab80e1ce76d7125c546006273e188 moved all clink config files
to vendor/ and therefore the files in config/ were not used and .history
was also moved to vendor/.

The new arrangement uses `--script` to load our own clink.lua file which in
turn as the first thing loads the original clink.lua file. The user defined
files are again in config/. As clink regenerates these files, if they are not
present, this also means that we don't need to include the settings file
in the git repo and therefore don't overwrite it on updates.

Closes: https://github.com/cmderdev/cmder/issues/793


JanSchulz:
Just a note: the current setup is broken as it places the clink files (mainly `history` and `settings`) in `vendor\` and not in `config\`.

JanSchulz:
ping?

MartiUK:
I need to figure this out, a lot of contradicting PRs and changes.

Stanzilla:
@MartiUK why not just merge this? Other PRs have to be rebased on that but this at least fixes the problem, no?

MartiUK:
Fair enough

Issue #772, opened 2015-12-14
CMDER_START should always be set to USERPROFILE unless explicitly set through /START parameter


MartiUK:
- [x] CmderLauncher.cpp: set CMDER_START to USERPROFILE unless set through /START param.
- [ ] Conemu.xml: Always use CMDER_START for new tasks.

JanSchulz:
cmder is also registered as `"C:\Users\jschulz\Dropbox\Programme\cmder\Cmder.exe" "%V"` but the startup requires `/START` https://github.com/cmderdev/cmder/blob/development/launcher/src/CmderLauncher.cpp#L290

JanSchulz:
Whatever, adding it also didn't make cmder start up in the directory I rightclicked.

MartiUK:
I haven't made these changes yet, but it should fix most issues regarding `/START`.

JanSchulz:
Ok, I think it's ok not to have /START in the context menu integration, but I think that the path argument is ignored in https://github.com/cmderdev/cmder/blob/development/launcher/src/CmderLauncher.cpp#L155 So in the moment it looks like the right click integration is broken? At least it doesn't work on my computer :-(

MartiUK:
The problem with cmder_here stems from the fact that the tasks do not use CMDER_START as an initial directory to work with.

JanSchulz:
What should the outcome be here? 

* `cmder /START <dir>` (right click "cmder here") -> should overwrite all settings?
* the tasks can specify it's own startup path 
* if a user sets `CMDER_START` -> not sure what priority that should have?

It seems for `PS`, the path gets applied to a registry key which is then used to set the `CWD`. For `cmd`, this registry key is ignored and `CMDER_START` env variable is used (which is not set by `cmder.exe`).  `shell` doesn't seem to change anything...

Unfortunately, the `init.bat` file always overwrites the current dir with the `"%HOME%"` if `CMDER_START` is not set.

IMO:
* if a path is set in `cmder /START <path>`, this path should overwrite everything
* if a user specifies a `CMDER_START` environment variable, it should be applied
* as a last measure, the set working dir (by the task) should be taken last.

* all tasks should set a working directory (which they do? `-new_console:d:%USERPROFILE%`)

IMO this could be done by letting `cmder.exe` setting/overwriting `CMDER_START` if  `path` is set. And get that implemented in all startup scripts.

JanSchulz:
[...]

PR #798, opened 2016-01-11
Make "cmder here" work again

JanSchulz:
The new system works by setting "CMDER_START" if a cmder.exe gets a
startup path. The rest is then implemented in the profile.ps1 and init.bat.
The new system works as follows:

1.If a startup path is given, cmder.exe sets this as `CMDER_START` and the
  startup scripts change to that directory.
2.If the usere sets a `CMDER_START` environment variable and this is not
  overwritten by cmder.exe, this variable is taken as the startup
  directory by the startup scripts.
3.If not such path is give, the conemu startup dirs is the winner, as the
  startup scripts don't change anything.

Closes: https://github.com/cmderdev/cmder/issues/772

### DIFF
--- a/launcher/src/CmderLauncher.cpp
+++ b/launcher/src/CmderLauncher.cpp
@@ -63,10 +63,12 @@ optpair GetOption()
 
 	if (argc == 1)
 	{
+		// no commandline argument...
 		pair = optpair(L"/START", L"");
 	}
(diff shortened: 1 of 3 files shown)


JanSchulz:
This doesn't yet work with bash... But on the other hand, on my system there isn't even a startup script run for bash.

Issue #203, opened 2014-06-02
Add an option to have cmder start in the current directory


pimterry:
I typically want a command line in the current folder I'm viewing in windows explorer. With cmd, I just type `cmd` into the address bar, and it opens in the current folder.

If I type `cmder` however, it opens in my home directory. I'm currently getting around this by renaming the cmder executable to cmder-raw, and creating a cmder.bat like:

``` bat
cmder-raw %CD%
```

(open 'cmder-raw', with the current directory as an argument, which apparently opens cmder there)

This does exactly what I want, but it's a bit hacky, and it results in a very very brief flash on non-cmder console window appearing, before cmder opens. It'd be great if cmder could add an option to act like this natively.

futurechan:
:+1: I would love the ability to type ```cmder``` into the address bar of file explorer and have it open in that directory!

mrgrain:
":+1: I would love the ability to type cmder into the address bar of file explorer and have it open in that directory!"

me too!

vitr:
:+1:

dvlsg:
:+1: Agreed, I really miss this ability when working with cmder.

hstarorg:
Me too.

kjeske:
Just go to cmder settings, then Tasks tab and set {cmd} task full command to:

````
cmd /k "%ConEmuDir%\..\init.bat" -new_console:%__CD__%
````

JanSchulz:
If you register cmder, then `right click -> Cmder here` works that way. 

For a solution which works as you currently work: add a startup batch file (`c.bat`) in a directory which is in the normal path (not the enhanced one in cmder!), which calls cmder with `cmder %CD%` (or `C:\Users\jschulz\Dropbox\Programme\cmder\cmder.exe %CD%` in my case). Its a little flickery, as the batch file opens ina window, but cmder comes up right in the right dir.

[Note that both solutions need a recent development build as there was for some time a bug which prevented the "start with a path"-scenario; not sure if the last release had this bug or not]

hstarorg:
@kjeske Thank you, it's available.

onewhaleid:
This is not working for me. cmder just opens in my home directory. What am I doing wrong?

![cmder](https://user-images.githubusercontent.com/11261876/36889750-6f53e50a-1e4f-11e8-9a39-5fdcac4121ce.png)

daxgames:
[...]

PR #803, opened 2016-01-15
Set tasks to always use CMDER_START

### LINKED ISSUE #772: CMDER_START should always be set to USERPROFILE unless explicitly set through /START parameter

MartiUK:
CMDER_START should always be set as a result of either:
1. The user passes a directory to cmder.exe using `/START $DIR` OR
2. Sets CMDER_START as a default environment variable.
Fixes #772

### DIFF
--- a/launcher/src/CmderLauncher.cpp
+++ b/launcher/src/CmderLauncher.cpp
@@ -160,7 +160,7 @@ void StartCmder(std::wstring path, bool is_single_mode)
 	}
 	else
 	{
-		static char buff[MAX_PATH];
+		static wchar_t buff[MAX_PATH] = { 0 };
 		GetEnvironmentVariable(L"USER_PROFILE", buff, MAX_PATH);
 		SetEnvironmentVariable(L"CMDER_START", buff);
 	}
--- a/launcher/src/CmderLauncher.cpp
+++ b/launcher/src/CmderLauncher.cpp
@@ -153,17 +153,15 @@ void StartCmder(std::wstring path, bool is_single_mode)
 		swprintf_s(args, L"/Icon \"%s\" /Title Cmder", icoPath);
 	}
 
-		SetEnvironmentVariable(L"CMDER_ROOT", exeDir);
-		if (!streqi(path.c_str(), L""))
-		{
-			SetEnvironmentVariable(L"CMDER_START", path.c_str());
-		}
-
-		// Send out the Settings Changed message - Once using ANSII...
-		//SendMessageTimeout(HWND_BROADCAST, WM_SETTINGCHANGE, 0, (LPARAM)"Environment", SMTO_ABORTIFHUNG, 5000, NULL);
-
-		// ...and once using UniCode (because Windows 8 likes it that way).
-		//SendMessageTimeout(HWND_BROADCAST, WM_SETTINGCHANGE, 0, (LPARAM) L"Environment", SMTO_ABORTIFHUNG, 5000, NULL);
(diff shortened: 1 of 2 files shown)


JanSchulz:
I don't understand the need for this. With https://github.com/cmderdev/cmder/pull/798, there are a few ways to startup:

* cmder here in windows explorer: will call `cmder.exe <path>` and `cmder.exe` will set `CMDER_START` which will be picked up by init.bat or profile.ps1 and change to that dir
* if a user specifies a `CMDER_START` this will take the same path as long as `cmder.exe` isn't called with a path/ `/START` (e.g. by "cmder here" or a commandline startup)
* if no path is supplied (e.g. by starting it up by a shortcut or plain `cmder.exe` in the shell), the current task definition will ensure that the shells are started in `USERPROFILE` which IMO are exactly the right dir to start in.

So IMO we are ok with the "start in the right dir" problem. Or what is still missing (-> whats the to be solved problem in #772)?

JanSchulz:
Just to make that clear: the above is currently happening for PS and cmd, but not for bash tasks. To make that happen as well, this could be applied:

```diff
diff --git a/vendor/cmder.sh b/vendor/cmder.sh
index 4d1da3a..9e21e31 100644
--- a/vendor/cmder.sh
+++ b/vendor/cmder.sh
@@ -51,3 +51,8 @@ fi
 if [ -f "${HOME}/.bashrc" ] ; then
     . "${HOME}/.bashrc"
 fi
+
+if [ "$CMDER_START" == "" ] ; then
+    CMDER_START="$(cygpath -u "$CMDER_START")"
+    cd "${CMDER_START}"
+fi
```

MartiUK:
I've reverted the config changes, although I'll keep the launcher changes as they resolve some issues where CMDER_START isn't set in some instances.

launcher/src/CmderLauncher.cpp
```diff
@@ -153,17 +155,26 @@ void StartCmder(std::wstring path, bool is_single_mode)
...
-		//SendMessageTimeout(HWND_BROADCAST, WM_SETTINGCHANGE, 0, (LPARAM)"Environment", SMTO_ABORTIFHUNG, 5000, NULL);
-
-		// ...and once using UniCode (because Windows 8 likes it that way).
-		//SendMessageTimeout(HWND_BROADCAST, WM_SETTINGCHANGE, 0, (LPARAM) L"Environment", SMTO_ABORTIFHUNG, 5000, NULL);
+	}
+	else
+	{
+		wchar_t* homeProfile = 0;
+		SHGetKnownFolderPath(FOLDERID_Profile, 0, NULL, &homeProfile);
+		if (!SetEnvironmentVariable(L"CMDER_START", homeProfile)) {
```
JanSchulz: IM this is still a bad change: init.bat will now *always* set the path to Home if cmder isn't started with a path. This means that startup path in the conemu config is completely useless now :-(

This also means that you can't do what's asked for in https://github.com/cmderdev/cmder/issues/203

JanSchulz:
See inline comments: IMO always setting `CMDER_START` is wrong. what is the usecase for that?

launcher/src/CmderLauncher.cpp
MartiUK: Yep just realised :fearful:

PR #832, opened 2016-02-10
Fixed checkGit() in case of submodules

gpakosz:
Since Git 1.7.8, submodules don't contain a .git/ folder anymore.
Instead, the submodule directory is populated with a .git text file
which contains a git-dir: /path/to/superproject/git_dir/modules/name.

See,
https://github.com/git/git/blob/master/Documentation/RelNotes/1.7.8.txt#L109-L114

### DIFF
--- a/vendor/profile.ps1
+++ b/vendor/profile.ps1
@@ -15,7 +15,7 @@ try {
 }
 
 function checkGit($Path) {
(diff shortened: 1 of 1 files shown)


MartiUK:
Ah sorry, could you redo this PR into the development branch?

Thanks. :+1:

gpakosz:
#835

PR #837, opened 2016-02-13
Add appveyor batch to README

JanSchulz:
This is mainly to point people to newer builds of cmder and is therefore to be
merged into the master branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ So, you've experimented with cmder a little and want to give it a shot in a more
 
 1. Open a terminal as an Administrator
 2. Navigate to the directory you have placed Cmder
-3. Execute `.\cmder.exe /REGISTER ALL`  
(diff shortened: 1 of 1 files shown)


JanSchulz:
Note: currently the URL for the badge itself does not work because it seems that there are multiple projects with the name `cmder` and appveyor stupidly only works when the name is unique. 
http://www.appveyor.com/docs/status-badges#badges-for-projects-with-public-repositories-on-github-and-bitbucket

One can use a different URL (see doc link above), but for that @MartiUK needs to comment here with the proper project ID :-)

JanSchulz:
Updated with the right URL for the badge...

JanSchulz:
Some day I will learn the Difference between batch and badge...  Sorry and thanks for the merge!

PR #849, opened 2016-02-19
Plugin support

daxgames:
Added  support to extend cmder using plugins in the form of json files.


scripts/build.ps1
```diff
@@ -55,6 +55,10 @@ Param(
 
 # Dot source util functions into this scope
 . ".\utils.ps1"
+
+# We need these for add-cmderplugin.ps1
+copy ".\utils.ps1" ../bin
```
JanSchulz: Is it really needed to copy a utils files into the bin directory (which is as I understand it for user scripts) and not used it via  `CMDER_ROOT`?

bin/add-cmderplugin.ps1
```diff
@@ -0,0 +1,161 @@
+<#
```
JanSchulz: Again, I don't think this should be installed into `bin`, because bin (the same as config) is under the user's control and therefore changes by the user should not be overwritten, but installing into bin would do this.

On the other hand, I don't see a better dir... The only way would be a `vendor/cmder` dir, which the user either needs to call the script by a absolute name or to add that dir to the path...

daxgames: Why would a user change stuff in bin?  I put it in bin so it would be in 
the path.

JanSchulz: Ok, the readme doesn't say anything else than "it will be in path" and it already has the alias.bat in it... Still seems untidy (like the stuff in config until it got mostly reorganized), but ok... :-)

daxgames: Since alias.bat was already there I thought it the best place to put it and 
the supporting utils.ps1.

Hopefully some people will try it out and we can start hosting plug-in json 
files along side the main repo.

daxgames:
I would like to ultimately see if this can be expanded to adding msys2 bits to git for windows inside cmder.  Thoughts?

JanSchulz:
I really like the idea (I've several different portable software packages under `%CMDER_ROOT%\vendor`), but on the other hand also see it as a duplication of e.g. chocolatey.

Re Msys2: I don't think it's good to mess with the bundled stuff in git for windows (which you can already include in the path, so you get `ls`, `rm` and such stuff), but I see the value of an independently installed msys2...

bin/add-cmderplugin.ps1
```diff
@@ -0,0 +1,161 @@
...
+#>
+
+[CmdletBinding(SupportsShouldProcess=$true)]
+Param(
+    # CmdletBinding will give us;
+    # -verbose switch to turn on logging and
+    # -whatif switch to not actually make changes
+
+    # Path to the vendor configuration source file
+    [string]$PluginPath = "$env:CMDER_ROOT\vendor\sources-plugins.json",
```
JanSchulz: The "bad" thing is, that this gets installed, so updating a package is basically releasing a new version of cmder...

bin/add-cmderplugin.ps1
```diff
@@ -0,0 +1,161 @@
...
+
+if (test-path $posh_user_aliases) {
+  $posh_aliases = $posh_user_aliases
+} else {
+  $posh_aliases = $posh_user_profile
+}
+
+
+$cmder_user_profile = $($env:cmder_root + "\config\user-profile.cmd")
+$cmder_aliases = $($env:cmder_root + "\config\aliases")
```
JanSchulz: Maybe it makes sense to rename that as well?

bin/add-cmderplugin.ps1
```diff
@@ -0,0 +1,161 @@
...
+    # We do not care about the extensions/type of archive
+    $tempArchive = "tmp/$($s.name).tmp"
+    Delete-Existing $tempArchive
+
+    if ($force -eq $true) {
+      write-host $("Deleting existing install of the '" + $s.name + "' plugin...")
+      Delete-Existing $s.name
+    }
+
+    if ((test-path $s.name) -eq $false) {
```
JanSchulz: Shouldn't this also do some test for the version and only complain if the same (or a higher) version number is already installed?

bin/add-cmderplugin.ps1
daxgames: I don't understand your comment.  'add-cmder-plugin.ps1' should be a part of the core cmder.  It could be updated separately if necessary by simply copying the file and utils.ps1 from the source repo into an installed binary release.

bin/add-cmderplugin.ps1
daxgames: Rename what 'add-cmderplugin.ps1'? To what?  I think it is already pretty self explanatory what it is.

bin/add-cmderplugin.ps1
daxgames: That could be added I guess.  Right now it complains if the directory is already there and prompts the user to use the '-replace' argument to force replacement of an existing plugin and then exits.

scripts/build.ps1
daxgames: Yes it is.

bin/add-cmderplugin.ps1
JanSchulz: Sorry, I mean the `sources-plugins.json` file. If thats is included in the cmder installer, the included versions are then only up to day on the release day of cmder and later probably out of date :-(

-> IMO this needs a path on the cloud which can change much more frequently...

bin/add-cmderplugin.ps1
JanSchulz: The latter: `config\aliases`

daxgames:
Your probably right about not messing with the bundled got for Windows 
stuff.  I was just trying to come up with a way to do it without having a 
second environment installed as I do now.

bin/add-cmderplugin.ps1
daxgames: That is probably an artifact from my starting point on adding plugin support.  I started with cmder_root\scripts\build.ps1 and modified it to do what I wanted it to do.  That is actually an error and has no function at this time.  I should probably remove the default.

bin/add-cmderplugin.ps1
daxgames: Yeah I personally have the following files in config:

user-aliases.doskey (sourced from user-profile.cmd)
user-aliases.sh (sourced from user-profile.sh)
user-aliases.ps1 (sourced form user-profile.ps1)

To maintain some consistency with other files in the folder.

vladimir-kotikov:
Totally agree with @JanSchulz about 
> but on the other hand also see it as a duplication of e.g. chocolatey

What is the benefits of using this feature instead of chocolatey? (which basically does the same thisngs from user perspective - installs stuff, adds it to the `%PATH%`...). Personally i don't like the idea of adding functionality that could be easily achieved using other well-known and wide spread tools.

daxgames:
Chocolatey is not portable.  If you move/copy your cmder folder to another 
system you would need to rerun the chocolatey commands.

This is adding to the cmder/vendor folder so if the plugin is portable in 
that it stores settings relative to the installs !  binaries you are 
extending a portable install of cmder that can just be picked up and moved 
as a whole.

vladimir-kotikov:
Ok, so what is the usecase? If you have some environment, which you want to have synced on every machine you use, it's easier to set up it manually once than create a declarative way to install it when you need it.

From the other side, if you want an ability to reproduce the environment on any machine, then it might be a good approach, but in this case IMO it is still better to reuse choco, plus some scripts to bind installed packages to Cmder.

Also, @daxgames, how do you planning to maintain these plugins, roll out updates, etc.?

daxgames:
All good points.  I was just trying to make it easy to extend and configure Cmder with portable apps.  This does that.

As for maintaining plugins I, honestly I did not think about that.  I assume if someone makes a plug-in they would maintain it.  I personally don't plan on maintaining anyone else's plugins.

I work from several different machines and I like my env to be as close to identical on all as possible.  Carrying a flash drive or having an archive in the cloud, or even a git repo/archive full of scripts to build an eng seems to me the best/easiest way to achieve this.  I am not a big Chocolatey user, I never even considered using it alongside Cmder.  Its a good idea, maybe I should try it.

I like to code and I just saw an opportunity to streamline and automate something that I have wanted and seen some people ask about.

Maybe it shouldn't be a part of core Cmder For the reasons you stated.  It could be an add-on that people can use of they want to.  Either way it does not matter to me, it's above my pay grade.

daxgames:
What about combining the approached?  Cmder plugins that use Chocolatey to do the installl and the magic of the plug-in is the binding of that Chocolatey managed install to the cmder env.  

My goal was to create something re-useable by many without everyone having to do custom scripting to extend Cmder.

vladimir-kotikov:
@daxgames, sorry, I didn't want to criticize your approach so agressively.

I'd love to have the functionality to easily set up my most-used tools to work with cmder, the only thing that concerns me it that using this approach we'll end up having Cmder directory really huge, since every plugin will install all stuff into `vendor` directory.

MartiUK:
This is interesting, I think this should be separate from cmder but can be included after it's fleshed out more.

Being able to update and replace parts of cmder install in a way we know works would be very powerful for users.

vladimir-kotikov:
@MartiUK, just another idea about how this could be used by Cmder - `add-cmderplugin` might be a part of Cmder build/pack process - for example we could have Cmder core (the thing that called now Cmder mini) and additional functionality which is added to core/packed using this script.

MartiUK:
@vladimir-kotikov I had an idea for something similar a while back. See: https://github.com/cmderdev/vendors, no reason why not.

daxgames:
That is sort of what have me the idea.  It was out there but under utilized 
since there was no way to apply an add-on vendor.  Which is why I wrote one.

dragon788:
I'd love to see this merged and bringing in something like pacman from Msys2 or allow easily installing extra functionality to cmder with something like Babun's "pact" utility, (itself a wrapper/rename of a cygwin util). Being able to install minimal, and then pull in additional pieces as I have time/bandwidth/need would be amazing.

Jackbennett:
The community and Microsoft are already making that with psGet+chocolaty though. We won't do a better job on our own it's something we're all watching closely to tie in.

You were always supposed to be able to just add onto the `vendor\sources.json` and re-run build.ps1 yourself to pull everything for your own cmder. It sounds like that's the user story that needs improving.

PR #862, opened 2016-03-02
Enhanced alias.bat to allow file storage path

daxgames:
This PR changes/enhances bin/alias.bat:

1. bin/alias.bat now has more robust command line parsing.
2. bin/alias.bat now uses config/user-aliases.cmd as the default alias store, this is to make things consistent across all shells.
3. config/user-aliases.cmd is a self executing batch file/doskey macro store.

  ```
  ;= @echo off
  ;= rem Call DOSKEY and use this file as the macrofile
  ;= %SystemRoot%\system32\doskey /listsize=1000 /macrofile=%0%
  ;= rem In batch mode, jump to the end of the file
  ;= goto:eof
  ;= Add aliases below here
  e.=explorer .
  gl=git log --oneline --all --graph --decorate  $*
  ls=ls --show-control-chars -F --color $*
  pwd=cd
  clear=cls
  history=cat %CMDER_ROOT%\config\.history
  unalias=alias /d $1
  vi=vim $*
  ```
4. vendor/init.bat will update existing config/aliases to the new self executing config/user-aliases.cmd format and remove the old file.
5. bin/alias.bat has a new command line argument '/f' that allows specification of the file store for new aliases.
6. bin/alias.bat now displays 'ALL' loaded macros if executed with no arguments.
7. bin/alias.bat deletes aliases from memory and the file store when '/d' is specified.

### DIFF
--- a/bin/alias.bat
+++ b/bin/alias.bat
@@ -1,64 +1,131 @@
 @echo off
 
-set ALIASES=%CMDER_ROOT%\config\aliases
-setlocal
-:: handle quotes within command definition, e.g. quoted long file names
-set _x="%*"
-set _x=%_x:"=%
+
+if "%aliases%" == "" (
+  set ALIASES=%CMDER_ROOT%\config\user-aliases.cmd
+)
+
+setlocal enabledelayedexpansion
+
+if "%~1" == "" echo Use /? for help & echo. & goto :p_show
 
 :: check command usage
-if ["%_x%"] == [""] echo Use /? for help & echo. & goto :p_show
-if ["%1"] == ["/?"] goto:p_help
-if ["%1"] == ["/reload"] goto:p_reload
-:: /d flag for delete existing alias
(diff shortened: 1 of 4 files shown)


daxgames:
I just wanted to explain the use case for this PR to justify its inclusion in core CMder. 

I use Cmder with a team of people as a shared work env.  We actually have a shared config in a git repo that we overlay on top of a binary Cmder distro so all team members can contribute to the shared config bits via source control.  This PR is how we are able to manage the shared aliases while allowing users to have user personal aliases.  We create '%cmder_root%\config\profile.d\shared-aliases.cmd' and '%cmder_root%\config\profile.d\user-aliases.cmd', they execute in that order so the user has final say and can override any shared alias.

MartiUK:
Could you resync your branch, thanks.

daxgames:
Done.  Sorry for the delay.

Issue #853, opened 2016-02-23
path entry for <git>/cmd instead of <git>/bin? 


janschulz:
The current git installer adds `C:\Program Files(x86)\Git\cmd` (and not `C:\Program Files(x86)\Git\bin`) to the path. IMO we should also change our path entry. `/cmd` would also make the `start-ssh-agent.cmd` script available in bash, which could then added (inactive...) to the default `config/user-profile.sh`.

On the other hand, `/bin` contains bash.exe, which might be good to have. So maybe add both?

Another thing I'm not a real an of is that `C:\Program Files\Git\usr\bin` is added at the start of the path, so things like windows `find` are shadowed by the git equivalent:

```cmd
λ where find
C:\Program Files\Git\usr\bin\find.exe
C:\Windows\System32\find.exe
c:\Rtools\bin\find.exe
```

I'm not sure if that was intentional (https://github.com/cmderdev/cmder/commit/fc85915b2c7b879daa371567467ef4921b22bbf6). 

So, IMO the path which is set if git is found should be:

```cmd
set "PATH=%GIT_INSTALL_ROOT%\cmd;%PATH%;%GIT_INSTALL_ROOT%\bin;%GIT_INSTALL_ROOT%\usr\bin;%GIT_INSTALL_ROOT%\usr\share\vim\vim74;"
```
[Or remove the `\usr\bin` entry and add that as an example to the generated `user-startup.cmd`]

And the same in the bash/ps startup scripts...

JanSchulz:
Ok, I implemented that in https://github.com/cmderdev/cmder/pull/865

-> Unix commands are *always* added *to the end of the PATH*.

Stanzilla:
Closed in favor of https://github.com/cmderdev/cmder/pull/865

Issue #782, opened 2015-12-30
Should we really be preferring externally installed Git?


daxgames:
Should we be doing this in the vendor\init.bat:

```
:: Check if msysgit is installed
@if exist "%ProgramFiles%\Git" (
    set "GIT_INSTALL_ROOT=%ProgramFiles%\Git"
) else if exist "%ProgramFiles(x86)%\Git" (
    set "GIT_INSTALL_ROOT=%ProgramFiles(x86)%\Git"
) else if exist "%USERPROFILE%\AppData\Local\Programs\Git" (
    set "GIT_INSTALL_ROOT=%USERPROFILE%\AppData\Local\Programs\Git"
) else if exist "%CMDER_ROOT%\vendor\git-for-windows" (
    set "GIT_INSTALL_ROOT=%CMDER_ROOT%\vendor\git-for-windows"
)
```

and this in the vendor\profile.ps1:

```
try {
    # Check if git is on PATH, i.e. Git already installed on system
    Get-command -Name "git" -ErrorAction Stop >$null
} catch {
    $env:Path += $(";" + $env:CMDER_ROOT + "\vendor\git-for-windows\bin")
}
```

Both of these could lead to running a different version of Git in cmd.exe/powershell.exe shells vs. the bash.exe shell shipped with Cmder.

If you really think we should, the vendor\init.bat code should be altered to the below:

```
:: Maintain compatibility with externally installed git for windows 1.9.5
@if defined GIT_INSTALL_ROOT (
    if exist "%GIT_INSTALL_ROOT%\share\vim\vim74\vim.exe" (
        set "PATH=%GIT_INSTALL_ROOT%\share\vim\vim74;%PATH%"
    ) else if exist "%GIT_INSTALL_ROOT%\usr\bin\vim.exe" (
        set "PATH=%GIT_INSTALL_ROOT%\usr\bin;%PATH%"
    )
)

:: Add git to the path
if defined GIT_INSTALL_ROOT (
    set "PATH=%GIT_INSTALL_ROOT%\bin;%PATH%"
    :: define SVN_SSH so we can use git svn with ssh svn repositories
    if not defined SVN_SSH set "SVN_SSH=%GIT_INSTALL_ROOT:\=\\%\\bin\\ssh.exe"
)

```

daxgames:
if not then I suggest:

vendor\init.bat

```
if exist "%CMDER_ROOT%\vendor\git-for-windows" (
[...]

PR #865, opened 2016-03-02
better git path handling

janschulz:
This PR contains ~~three~~ two main commits

* "init.bat: check for git in path" implements the scheme from https://github.com/cmderdev/cmder/issues/782#issuecomment-169622265
* "init.bat: integrate the unix commands at the end of PATH" implements https://github.com/cmderdev/cmder/issues/853

### DIFF
--- a/vendor/init.bat
+++ b/vendor/init.bat
@@ -65,18 +65,23 @@ if not defined TERM set TERM=cygwin
         set test_dir=
     )
 )
+
 :: our last hope: our own git...
 :VENDORED_GIT
 @if exist "%CMDER_ROOT%\vendor\git-for-windows" (
     set "GIT_INSTALL_ROOT=%CMDER_ROOT%\vendor\git-for-windows"
+    rem add the minimal git commands to the front of the path
+    set "PATH=%GIT_INSTALL_ROOT%\cmd;%PATH%"
 ) else (
     goto :NO_GIT
 )
 
 :FOUND_GIT
 :: Add git to the path
-if defined GIT_INSTALL_ROOT (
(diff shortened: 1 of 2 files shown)


vendor/profile.ps1
```diff
@@ -41,6 +41,8 @@ try {
     # Check if git is on PATH, i.e. Git already installed on system
     Get-command -Name "git" -ErrorAction Stop >$null
 } catch {
+    $env:Path += $(";" + $env:CMDER_ROOT + "\vendor\git-for-windows\cmd")
+    # for bash.exe, which in the cmd version is found as <GIT>\usr\bin\bash.exe
     $env:Path += $(";" + $env:CMDER_ROOT + "\vendor\git-for-windows\bin")
```
JanSchulz: I'm actually not sure if this is a nice thing to keep the `...\bin` version: the problem is that if the user has added git to the path (via the git for windows installer), then bash is not found as the installer does not add bin to the path.

vendor/init.bat
```diff
@@ -95,9 +126,12 @@
 )
 
 @pushd "%CMDER_ROOT%\config\profile.d"
-for /f "usebackq" %%x in ( `dir /b *.bat *.cmd` ) do (
-  REM @echo Calling %CMDER_ROOT%\config\profile.d\%%x...
-  @call %%x
+@>nul 2>nul dir /b *.bat *.cmd && (
```
daxgames: Is this really necessary, it's confusing?  If there are no files in the profile.d folder then the for loop will not do anything.  You are just running ```dir /b *.bat *.cmd``` twice instead of once in the event there are files.  Why run it twice if once does the job?

daxgames:
I did add it to mine.

JanSchulz:
Updated and repushed

MartiUK:
Could you resync, your branch again? Thanks.

janschulz:
Updated, sorry for the long delay, seems I missed the rebase request.

PR #867, opened 2016-03-02
Various fixes for profile.d support

daxgames:


vendor/init.bat
```diff
@@ -97,7 +97,7 @@
 @pushd "%CMDER_ROOT%\config\profile.d"
 for /f "usebackq" %%x in ( `dir /b *.bat *.cmd` ) do (
   REM @echo Calling %CMDER_ROOT%\config\profile.d\%%x...
-  @call %%x
+  @call "%CMDER_ROOT%\config\profile.d\%%x"
```
daxgames: This is to account for other scripts that are in profile.d that MAY use 'cd' in them instead of pushd/popd and thus break this section of init.bat.

PR #873, opened 2016-03-06
Prefer use of first line @echo off vs. @ per line to turn off echo pe…

daxgames:
Prefer use of first line @echo off vs. @ per line to turn off echo per line. - Cleaner file, no side efects

### DIFF
--- a/vendor/init.bat
+++ b/vendor/init.bat
@@ -1,3 +1,5 @@
+@echo off
+
 :: Init Script for cmd.exe
 :: Created as part of cmder project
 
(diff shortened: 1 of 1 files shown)


JanSchulz:
Hm, interesting, I really wonder why this comment is in the start-ssh-agent script:

`@REM Do not use "echo off" to not affect any child calls.`

https://github.com/git-for-windows/MINGW-packages/blob/master/mingw-w64-git/start-ssh-agent.cmd#L1

"Calls in the script" vs "calls after the script exits and the cmd is available for user commands"?

JanSchulz:
Just to make sure:

running this script and then runing a batch file **without `echo off`** will echo the commands?

I just tried to simulate that by calling echo off in the normal cmder windows and then a script, and it didn't work:

```
c:\data\external\cmder (where_is_git)
λ test.cmd

where /Q git.exe

if errorlevel 1 goto :VENDORED_GIT

goto :EOF

c:\data\external\cmder (where_is_git)
λ echo off
c:\data\external\cmder (where_is_git)
λ test.cmd
c:\data\external\cmder (where_is_git)
λ
```

JanSchulz:
Ok, it seems this only affects the calls form the init.bat, not something which happens after:

```
c:\data\external\cmder (where_is_git)
λ cmd /k test2.bat

c:\data\external\cmder {git}{hg}
{lamb} ﻿where git
Der Befehl "﻿where" ist entweder falsch geschrieben oder
konnte nicht gefunden werden.

c:\data\external\cmder {git}{hg}
{lamb} echo off
ECHO ist ausgeschaltet (OFF).

c:\data\external\cmder {git}{hg}
{lamb} echo
ECHO ist eingeschaltet (ON).

c:\data\external\cmder {git}{hg}
{lamb}
```

daxgames:
Correct.  It turns off echo for the duration of the current script. 

I did the exact same testing.

vendor/init.bat
```diff
@@ -80,39 +82,39 @@
...
+    echo Creating user startup file: "%CMDER_ROOT%\config\user-profile.cmd"
     (
-    @echo :: use this file to run your own startup commands
-    @echo :: use @ in front of the command to prevent printing the command
-    @echo.
-    @echo :: @call "%%GIT_INSTALL_ROOT%%/cmd/start-ssh-agent.cmd
-    @echo :: @set PATH=%%CMDER_ROOT%%\vendor\whatever;%%PATH%%
-    @echo.
+    echo :: use this file to run your own startup commands
+    echo :: use  in front of the command to prevent printing the command
```
bondz: This line should be out.

Issue #568, opened 2015-07-31
WARNING: Enable-GitColors is Obsolete and will be removed in a future version of posh-git.


eeree:
Recently `posh-git` has been updated and some features started to be deprecated. One of them is `Enable-GitColors`. From now on this feature is ON by default and users, who run the command `Install-Module posh-git` will get the following error:

```
WARNING: Enable-GitColors is Obsolete and will be removed in a future version of posh-git.
```
![ixp3wl5](https://cloud.githubusercontent.com/assets/1375008/9015782/a1fbea38-37cb-11e5-832d-cf56cce6139f.png)

eeree:
Attached fix does NOT remove colors from git repositories. Newer version of `posh-git` has simply made this option _default_.
![2015-07-31 21_52_34-cmder](https://cloud.githubusercontent.com/assets/1375008/9016098/7c3a0dd6-37ce-11e5-870d-bb7b4c2941b1.png)

puradox:
If you're tired of the warning, simply comment out `Enable-GitColors` on line 54 in `vendor/profile.ps1` (location may vary on release).

```PowerShell
# Load special features come from posh-git
if ($gitStatus) {
    #Enable-GitColors
    Start-SshAgent -Quiet
}
```

eeree:
@puradox True. This is why I suggested a change in a pull request, to remove this unnecessary line of code. It's not used anymore.

eeree:
To clear things up. In [this commit!](https://github.com/dahlbyk/posh-git/commit/4e778e2480f584e039b47f0db1c020a197ee9e8f) `Enable-GitColors` has been removed. As for now it's just a dummy function - colors are enabled by default (as you see on a screenshot above).

maximejobin:
Awesome. Commenting `Enable-GitColors` worked.

PR #878, opened 2016-03-09
Remove Enable-GitColors from profile.ps1

jamiepenney:
This is a default in Posh-Git now. It complains about it being obsolete, and it will be removed in future versions of Posh-Git

See https://github.com/cmderdev/cmder/issues/568


Stanzilla:
This is already gone in the next version, see the development branch.

jamiepenney:
My bad - must have missed that PR in the activity feed of the issue!

PR #892, opened 2016-03-21
fixed - not running user-aliases.cmd if aliases variable is overridde…

daxgames:
fixed - profile.d overridden %aliases% store not running config/user-aliases.cmd

### DIFF
--- a/vendor/init.bat
+++ b/vendor/init.bat
@@ -29,7 +29,7 @@ if "%PROCESSOR_ARCHITECTURE%"=="x86" (
 if not exist "%CMDER_ROOT%\config\settings" (
     echo Generating clink initial settings in %CMDER_ROOT%\config\settings
     echo Additional *.lua files in %CMDER_ROOT%\config are loaded on startup.
-) 
+)
 
 :: Run clink
 "%CMDER_ROOT%\vendor\clink\clink_x%architecture%.exe" inject --quiet --profile "%CMDER_ROOT%\config" --scripts "%CMDER_ROOT%\vendor"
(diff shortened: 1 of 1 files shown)


daxgames:
Did you already have a user-aliases.cmd? or did you rename aliases manually to user-aliases.cmd?

Stanzilla:
Nope, only the old `aliases` file.

daxgames:
The logic for the update prior to my pull requests today is below

```
if not exist "%user-aliases%" (
   echo Creating intial user-aliases store in "%user-aliases%"...
   copy "%CMDER_ROOT%\vendor\user-aliases.cmd.example" "%user-aliases%"
)

:: Update old 'user-aliases' to new self executing 'user-aliases.cmd'
if exist "%CMDER_ROOT%\config\aliases" (
  echo Updating old "%CMDER_ROOT%\config\aliases" to new format...
  type "%CMDER_ROOT%\config\aliases" >> "%user-aliases%" && del "%CMDER_ROOT%\config\aliases"
)
```
So if all you had was the old aliases file and no pre-existing  'user-aliases.cmd' a new 'user-aliases.cmd' SHOULD have been created from the new 'vendor\user-aliases.cmd.example' which contains the self extracting logic.  Then if the 'aliases' file exists it is appended to the new 'user-aliases.cmd' and 'aliases is deleted at that point.

I don't see a flaw unless your 'vendor\user-aliases.cmd.example' does not have the self extracting loigic and I can't explain that.

I did a pull request today that makes the process a little more robust.  Please let me know if you figure out why it does not work.

Stanzilla:
Yup figured it out, I didn't have the example file at all, my fault!

daxgames:
Cool, glad you figured it out.

PR #916, opened 2016-04-09
Fix missing "\" when building dirpath to psmodules

liftir:
Join-Path won't add "\"'s without the use of a delimeter.  In this case there is no delimiter and as a result the script will fail when it tries to look up $CmderModulePath.

### DIFF
--- a/vendor/profile.ps1
+++ b/vendor/profile.ps1
@@ -18,7 +18,7 @@ if(!$PSScriptRoot) {
 }
 
 # Add Cmder modules directory to the autoload path.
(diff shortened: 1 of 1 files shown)


Stanzilla:
@liftir @Jackbennett revert this then?

Jackbennett:
I just can't find an example this fixes but join-path won't correct too many `\`'s so in my opinion makes it more fragile to its inputs.

Jackbennett:
@Stanzilla  Might as well stay since it's on but it doesn't do anything. As long as nobody tries to think it needs escaping as "\/psmodules/" will break.

Stanzilla:
ok

PR #917, opened 2016-04-09
Fix missing "/" when joining paths #2

liftir:
Woops missed a  join-path.  Added another "\"

### DIFF
--- a/vendor/profile.ps1
+++ b/vendor/profile.ps1
@@ -18,7 +18,7 @@ if(!$PSScriptRoot) {
 }
 
 # Add Cmder modules directory to the autoload path.
-$CmderModulePath = Join-path $PSScriptRoot "psmodules/"
+$CmderModulePath = Join-path $PSScriptRoot "\psmodules/"
 
 if( -not $env:PSModulePath.Contains($CmderModulePath) ){
     $env:PSModulePath = $env:PSModulePath.Insert(0, "$CmderModulePath;")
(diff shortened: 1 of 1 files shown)


liftir:
me no know git

Jackbennett:
What problem does this actually fix? `join-path` takes care of exactly this and my testing I can't seem to trip it up. Did you run into something?

liftir:
join-path will only add the "\" between paths if you use a delimiter.  from the help:

Synopsis
    Combines a path and a child path into a single path. The provider supplies the path delimiters.

Syntax
    Join-Path [-Path] <String[]> [-ChildPath] <String> [-Credential <PSCredential>] [-InformationAction ] [-InformationVariable <System.String>] [-Resolve ] [<CommonParameters>]


Inputs
    System.String
    You can pipe a string that contains a path to Join-Path.

Outputs
    System.String
    Join-Path returns a string that contains the resulting path.

Notes
    The cmdlets that contain the Path noun (the Path cmdlets) manipulate path names and return the names in a concise format that all Windows PowerShell providers can interpret. They are designed for use in programs and scripts where you want to display all or part of a path name in a particular format. Use them like you would use Dirname, Normpath, Realpath, Join, or other path manipulators.
    

Examples
    -------------------------- EXAMPLE 1 --------------------------
    PS C:\>join-path -path c:\win* -childpath System*
    
    This command uses Join-Path to combine the "c:\Win*" path with the "System*" child path. The Windows PowerShell file system provider, FileSystem joins the path and adds the "\" delimiter.

Jackbennett:
I'm sorry @liftir but I can't quite pick out part of the help which states it needs to have a delimiter. 

From testing, if anything you want to leave out your own delimiters because PS seems to only add them, not fix an incorrect path with too many delimiters. see example 5.

```
join-path "a" "b"
join-path "a" "b/"
join-path "a.txt" "b"
join-path "\a\" "\b\"
join-path "c:/a//" "//b/"
join-path "a/" "/b/"

a\b
a\b\
a.txt\b
\a\b\
c:\a\\\b\
a\b\
```

liftir:
Hmm it was throwing an error for me until I added those extra "'\"'s.  $CmderModulePath would be NULL.  Now that the environment has been setup, it's working fine without the additions.

Stanzilla:
Revert or not revert then?

Issue #420, opened 2015-03-12
Single mode does not set current directory


DoCode:
When open cmder([cmder launcher exe](https://github.com/bliker/cmder/blob/development/launcher/src/CmderLauncher.cpp#L115) from development branch) in single mode from explorer context menu,
the `CMDER_START` environment variable is not overwritten when the second tab is started.

MartiUK:
Just to be sure, are you running an exe you compiled yourself or the one included in the repo?

DoCode:
I am running an self compiled exe from the current development branch who has the `/single` cli-switch.

rainyday:
1.3.2 release notes claim this  issue was fixed but that does not appear to be the case. running `cmder /single C:\path` still does not set the path correctly in the new tab

DoCode:
I don't know why me PR that resolve this issues with `/single` and `Cmder startup directory` is gone!?  
See it [here](https://github.com/cmderdev/cmder/pull/673).

Stanzilla:
@DoCode that is weird? It shows in the history? https://github.com/cmderdev/cmder/commits/master/launcher/src/CmderLauncher.cpp

Stanzilla:
no idea what happened, sorry :( Maybe @daxgames @MartiUK know. 

Care to redo the PR then?

Starli0n:
I am facing the same issue here but only when I had launched an instance of Cmder before.

If I execute `Cmder.exe /SINGLE D:\` when there is no `Cmder` launched, it works and it will work it I run again this command.

However, if I had launched `Cmder` manually before the same command would not work and the default path would be my user path.

I tweaked the options of `Running As Admin` and tried all kind of combinations but I did not figure out how to solve the issue.

I do not know whether those issues are linked.

DoCode:
@Starli0n please try my [PR](https://github.com/cmderdev/cmder/commit/a57c11e171f97eb8cc6a10b61053731d2940316a). This worked for me since Sep 4, 2015!

DoCode:
@Stanzilla, how can we close this issue for all times? I use my [PR](https://github.com/cmderdev/cmder/commit/a57c11e171f97eb8cc6a10b61053731d2940316a) since Sep 4, 2015 and it works like a Charme with all updates from ConEmu...

Starli0n:
@DoCode Thanks you, that's nice but I do not intend to compile my own version of Cmder (at least for now)

So please integrate his PR

Starli0n:
[...]

PR #921, opened 2016-04-13
use /dir Switch instead of CMDER_START

david-driscoll:
So I've been trying to use `/single <path>` to open a new console, in an existing cmder instance.  And currently it's not working.  At first I thought it was because the environment variable wasn't be set (it was).

I think I've nailed down the scenario that's the problem... I unfortunately don't have VS2015 and and friends installed on this machine, so I will review this later tonight as well.

When opening a new instance of Cmder `Cmder.exe /single C:\`.  `CMDER_START` is set properly, and the folder is correct.

When trying to open another instance of Cmder `Cmder.exe /single C:\` the directory always defaults to   `%USERPROFILE%`.  Now this is because the directory is set for the new console from `ConEmu`.  In this scenario what appears to be happening is that the Environment Variable `CMDER_START` on the original `Cmder` instance (which is a wrapper for `ConEmu`?) will not be changed or set when `/single` is called.

I believe this value can be set via the `/dir` switch on ConEmu itself, so this change is an attempt to do that.

As I said I don't have the tool chain installed on this machine to really test the changes, and if this change is completely off base feel free to let me know.

### DIFF
--- a/config/ConEmu.xml
+++ b/config/ConEmu.xml
@@ -488,7 +488,7 @@
 				<key name="Task1" modified="2015-11-24 14:49:10" build="151119">
 					<value name="Name" type="string" data="{cmd::Cmder as Admin}"/>
 					<value name="GuiArgs" type="string" data=" /icon &quot;%CMDER_ROOT%\icons\cmder.ico&quot;"/>
-					<value name="Cmd1" type="string" data="*cmd /k &quot;%ConEmuDir%\..\init.bat&quot;  -new_console:d:%USERPROFILE%"/>
+					<value name="Cmd1" type="string" data="*cmd /k &quot;%ConEmuDir%\..\init.bat&quot;"/>
 					<value name="Active" type="dword" data="00000000"/>
 					<value name="Count" type="dword" data="00000001"/>
 					<value name="Hotkey" type="dword" data="00000000"/>
@@ -497,7 +497,7 @@
 				<key name="Task2" modified="2015-11-24 14:49:10" build="151119">
 					<value name="Name" type="string" data="{cmd::Cmder}"/>
 					<value name="GuiArgs" type="string" data=" /icon &quot;%CMDER_ROOT%\icons\cmder.ico&quot;"/>
-					<value name="Cmd1" type="string" data="cmd /k &quot;%ConEmuDir%\..\init.bat&quot;  -new_console:d:%USERPROFILE%"/>
+					<value name="Cmd1" type="string" data="cmd /k &quot;%ConEmuDir%\..\init.bat&quot;"/>
 					<value name="Active" type="dword" data="00000000"/>
 					<value name="Count" type="dword" data="00000001"/>
 					<value name="Hotkey" type="dword" data="00000000"/>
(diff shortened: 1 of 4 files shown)


Stanzilla:
Can probably loop in @Maximus5 to confirm that

Maximus5:
I have no idea, how `/single` in cmder is implemented.

Empty switch `-new_console` in Tasks is useless.

Using `%CMDER_ROOT%` in profile initialization scripts is bad idea - they would fail if user starts just a `ConEmu.exe`.

Stanzilla:
Okay, thanks for your input!

janschulz:
> When trying to open another instance of Cmder Cmder.exe /single C:\ the directory always defaults to %USERPROFILE%. Now this is because the directory is set for the new console from ConEmu. In this scenario what appears to be happening is that the Environment Variable CMDER_START on the original Cmder instance (which is a wrapper for ConEmu?) will not be changed or set when /single is called.

Somehow this sounds like a bug and maybe it is easier to find out why in that case CMDER_START is not set?

david-driscoll:
Well the `CMDER_START` environment variable belonges to the other process (the first copy of Cmder that was started).  So when you say `cmder /single <path>`

So the basic example of the issue, from any command prompt (just plain old `cmd`), with Cmder closed.

Run:

```
cmder /single c:\
cmder /single c:\windows
cmder /single
```

In all three cases the default directory will be `C:\` because, the original, and `single` host process always has the same default environment variable for `CMDER_START` set.  You can repeat it doing `/single` first, and then the default will always be `%USERPROFILE%`.

With these changes, now that I've dusted off my C++ (It's been a long while forgive me!).  All of the above commands open as expected, to the correct directory.

Let me know if you need more information or details!

janschulz:
Ok, do I understand that right, that if I start cmder via an icon, the conemu process has no `CMDER_START` env variable. When I then start another `cmder /single C:`, this sets `CMDER_START` in the current process and starts a new conemu (which also still has that set), but this will find another conemu already running and passes control to that instance, but that instance has no `CMDER_START` set, so the started new child cmd will also not have it?

If so we have to look for another way to do this...

The normal startup (as far as I thought when I worked on the launcher code for `CMDER_START`) should have all this properties (in decreasing priority: first condition satisfied will win):

* If the user specifies any dir to `cmder <path>`, it should startup in that path
* if the user has specified `CMDER_START` as env variable, it should startup there
* If the user sets a startup parameter in the ConEmu task spec, it should startup there.
* The normal startup without any parameter and nothing set should startup in %USERPROFILE%

>  I unfortunately don't have VS2015 and and friends installed on this machine, so I will review this later tonight as well.

You can download the built cmder form appveyor -> just follow the link in the "checks" to appveyor and there to the artefacts.

janschulz:
> Using %CMDER_ROOT% in profile initialization scripts is bad idea - they would fail if user starts just a ConEmu.exe.

How should we work around that? The only thing I can think of is using `ConEmuDir` env var to compute CMDER_ROOT in each startup script...?

david-driscoll:
I have VS2015 on my home laptop and surface, just not on my work machine.  So I was able to get it it work.

With this PR currently it does:

* If the user specifies any dir to cmder <path>, it should startup in that path
* The normal startup without any parameter and nothing set should startup in %USERPROFILE%

> If the user sets a startup parameter in the ConEmu task spec, it should startup there.
I'm not sure if their task spec `/dir` would override the one provided by `Cmder`



I could add a step before `USERPOFILE` that looks for `CMDER_START`.  I just wonder how much value is the environment variable `CMDER_START`?

janschulz:
>  I just wonder how much value is the environment variable CMDER_START?

I wouldn't mind if it is gone... But others have to decide this.

What makes this issue interesting is that if both CMDER_START and CMDER_ROOT would go away, there is IMO not much reason to keep the launcher around compared to using conemu directly... The only thing would be the copying of conemu config files on different machines...

MartiUK:
@david-driscoll Can you resolve the conflicts, I'll review and merge when that's done.

Jackbennett:
Only done some cursory reading (because I'm tired of looking at the word "single" haha) but this sounds like it might be worth looking through #420 and friends to make sure various things are accounted for.

Stanzilla:
continue here https://github.com/cmderdev/cmder/pull/1609

Issue #937, opened 2016-05-05
"\cmder\config\settings was unexpected at this time" and {lamb} is shown instead of lambda symbol


ElonLiu:
This happened since I downloaded the newest version and copied my old `aliases`, `ConEmu.xml` and `settings` file into the new config/ folder. I thought maybe I shouldn't have copied the `settings` file. But even though I deleted it, things are still the same.
Strangely, if I move the whole cmder to a different location, everything goes well. But it will not work in my previous location. Should I clean anything and where?

ElonLiu:
Sorry. This issue seems similar to issue #935 and I fixed it according to @denisz1 

In addition to @denisz1's modifications, another modification may be as well needed: (Line 31)

Original:
`echo Additional *.lua files in %CMDER_ROOT%\config are loaded on startup.`

Should be:
`echo Additional *.lua files in "%CMDER_ROOT%\config" are loaded on startup.`

janschulz:
Can you try this here: https://ci.appveyor.com/project/MartiUK/cmder/build/1.0.267-development/artifacts

Issue #935, opened 2016-05-02
init.bat fails and shows {lamb} and {git}


denisz1:
For the latest pre release
The `init.bat` fails, and the prompt is all wrong

File fails here:

    if not exist "%CMDER_ROOT%\config\settings" (
        echo Generating clink initial settings in %CMDER_ROOT%\config\settings
        echo Additional *.lua files in %CMDER_ROOT%\config are loaded on startup.
    ) 

There is another location in the bat file with the same problem.

janschulz:
What ist the error message?

denisz1:
**UPDATE:**

I fixed it by adding _quotation marks_. Do the following:

Change this:

        echo Generating clink initial settings in %CMDER_ROOT%\config\settings
        echo Additional *.lua files in %CMDER_ROOT%\config are loaded on startup.

To this:

        echo Generating clink initial settings in "%CMDER_ROOT%\config\settings"
        echo Additional *.lua files in "%CMDER_ROOT%\config" are loaded on startup.

And change this:

        echo Creating intial aliases in %CMDER_ROOT%\config\aliases

To this:

        echo Creating intial aliases in "%CMDER_ROOT%\config\aliases"

Then it installs and runs as expected.

janschulz:
Ok, can you please still post the error message and also the path where you installed cmder in? I suspect that it as something to do with spaces in your path, but actually can't really understand the problem because it is just a echoed variable and not something which is used programmatically (like the `if not exist "%CMDER_ROOT%\config\settings" (` line).

denisz1:
There was no error. It didn't run far enough to install clink. So like others have reported, the prompt said {lamb} instead of the lambda sign, and the {git} substitutions were not made. And all paths were correct. It was a brand new install.

Others have reported that it was because of an old version of clink, but that's not it. It was not running clink, because the batch file failed at those lines. Makes no sense, btu there you go. 

Win10 x64 BTW.

janschulz:
> There was no error. 

No error message at all when you opened a new cmder window? Also: what is the full path to your cmder install location (e.g. in my case: `C:\Users\jschulz\Dropbox\Programme\cmder\Cmder.exe`)?

denisz1:
No, no error. But clink didn't run, so the macros were not expanded, and the shell was not "upgraded" in the cmder style.

`c:\Program Files (x86)\foo\bar\cmder\Cmder.exe`

denisz1:
[...]

PR #939, opened 2016-05-05
Add quotes around all variables

janschulz:
This should prevent errors like
https://github.com/cmderdev/cmder/issues/935
https://github.com/cmderdev/cmder/issues/937

I've no idea why this happens, but it shouldn't hurt anyway...


janschulz:
This is against master... :-/

PR #940, opened 2016-05-05
Add quotes around all variables

janschulz:
This should prevent errors like
https://github.com/cmderdev/cmder/issues/935
https://github.com/cmderdev/cmder/issues/937

I've no idea why this happens, but it shouldn't hurt anyway...

### DIFF
--- a/vendor/init.bat
+++ b/vendor/init.bat
@@ -12,7 +12,7 @@ if not defined CMDER_ROOT (
 )
 
 :: Remove trailing '\'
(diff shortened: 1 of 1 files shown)


janschulz:
Now against branch development, for the one against master see https://github.com/cmderdev/cmder/pull/939

janschulz:
I tested both this version and the previous development version ins a `c:\temp\test test\` dir, and both were starting up fine... No idea what the problem was, but if this fixes it, and it works for the rest, too...

PR #941, opened 2016-05-06
More variable quoting in init.bat

janschulz:
There were problems when a path contained a parenthesis like `C:\temp\test (test)\`

As a precaution, quote all variables when they are used in echo or set.

### DIFF
--- a/vendor/init.bat
+++ b/vendor/init.bat
@@ -8,7 +8,7 @@
 
 :: Find root dir
 if not defined CMDER_ROOT (
-    for /f "delims=" %%i in ("%ConEmuDir%\..\..") do set CMDER_ROOT=%%~fi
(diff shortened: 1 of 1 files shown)


vendor/init.bat
```diff
@@ -8,7 +8,7 @@
 
 :: Find root dir
 if not defined CMDER_ROOT (
-    for /f "delims=" %%i in ("%ConEmuDir%\..\..") do set CMDER_ROOT=%%~fi
+    for /f "delims=" %%i in ("%ConEmuDir%\..\..") do set "CMDER_ROOT=%%~fi"
```
BrainCrumbz: Are you sure you want to wrap the whole assignment within double quotes? The left side of assignment is basically declaring the variable/identifier, and should not need quotes. The right side instead is what might be affected by double quotes. Just to be sure, have you checked if it works jut with `..... do set CMDER_ROOT="%%~fi"`

Issue #950, opened 2016-05-12
Feature: add some extra prompt-tuning hooks to profile.ps1 from user-profile.ps1


centur:
I'd like to suggest a feature for Cmder Powershell story:

`profile.ps1` is doing prompt tuning by defining `function global:prompt` which is a normal powershell way of tuning prompts. Unfortunately there is no extension points added to a current fine-tuned prompt.

So if user wants to add extra customization or some hooks to their prompt via `user-profile.ps1` (like persistent history and other neat tricks) - it's impossible as the only thing one can do is to override `global:prompt` and loose all cmder neat customization.

I'd like to provide few simple function hooks - which user can define in `user-profile.ps1` but they will be executed in each prompt call.

I'm using this technique for awhile and found it pretty convenient, as well as nice performing - 
Invoke-Expression doesn't add much overhead compared to the direct function call. I quickly smoke measured it with Measure-Command before PR and saw that it adds about 
0.001 to 0.003 secs compared to direct invocation of the function which measured, with an empty function, to ~0.008)

So in overall this technique adds from 0.003 to 0.009 sec overhead (3 to 9 milliseconds) which is unnoticeable by a normal user.

Jackbennett:
Referenced on PR #952 to focus discussion into one place where the change may be made.

PR #951, opened 2016-05-12
feat: add prompt-tuning hooks to Cmder default customization to allow…

centur:
This is an implementation of #950 feature to allow user define powershell prompt customization in user-profile.ps1, not by editing vendor/profile.ps1


Jackbennett:
I don't want to step on any toes but can we move discussion onto #952

Issue #528, opened 2015-06-10
Powershell tab title issue


edclement:
Under Settings>Main>Tabs, if you use a tab template of "%f" (folder) or "%d" (current directory) for the tab name and use powershell as your prompt, the tab title won't actually update with the current directory. If you do use cmd.exe as your prompt instead, the tab title does update properly. I managed to figure out the fix for this. Adding this line:
`& ConEmuC64 -StoreCWD "$dir"`
...to the `cmder\vendor\profile.ps1` powershell profile as the last statement of the `function global:prompt {` function does allow for the tab title to update properly on directory navigation.
So the full function looks like:
```
function global:prompt {
    $realLASTEXITCODE = $LASTEXITCODE
    $Host.UI.RawUI.ForegroundColor = "White"
    Write-Host $pwd.ProviderPath -NoNewLine -ForegroundColor Green
    if($gitStatus){
        checkGit($pwd.ProviderPath)
    }
    $global:LASTEXITCODE = $realLASTEXITCODE
    Write-Host "`nλ" -NoNewLine -ForegroundColor "DarkGray"
    & ConEmuC64 -StoreCWD "$dir"
    return " "
}
```

Can one of the project maintainers add this one liner?

Jackbennett:
That line `& ConEmuC64 -StoreCWD "$dir"` doesn't do anything for me even `& ConEmuC64 -StoreCWD "test"`

I don't see $dir defined in your post I'm not even sure how yours is working.

The line I get it to work with is: `conemuC -GuiMacro Rename 0 'tabname'` but putting that into the prompt slows it down by a few tenths which I don't find acceptable. I tried running it as a background task and it's just as slow. 

I added the host process title in though so at least windows knows that process's name by it's directory. You see this in the start aero peak list of cmders open tabs

PR #952, opened 2016-05-12
Custom prompt hooks protected from later overwriting

Jackbennett:
In reference to issue #950 about PR #951 I have the following suggestion.

@centur I like your suggestion, I'm just uneasy out by the idea of running any function that is called `User-PrePrompt`, `User-Prompt`, `User-PostPrompt` as the prompt reloads. After trying that out I wasn't enthused to find the following idea has some fun effects.

`function Write-Host {'Hello Boys, I'm Back'}`

So with that, I propose: 
- Add a pre and post function hook around the Cmder prompt.
- Specify the cmder prompt as a function that could be replaced by a user.
- Write a friendly message when the user profile template is created.
- Create the user profile with cmder prompt hooks ready to use.

Core functions have been explicitly called from their namespace like Microsoft.PowerShell.Utility\Write-Host to try and prevent clobbering with later defined functions.

User supplied functions are variables in as script blocks, created as the session runs the profile script. By creating them as constants these function names cannot be re-declared for the duration of the process.

Since the prompt function already exists by this time, set the readOnly flag so to re-declare the prompt requires the use of -force.

It is hoped these changes limit what could be the risk of any script redefining functions that are called automatically without user intent or input.

### DIFF
--- a/vendor/profile.ps1
+++ b/vendor/profile.ps1
@@ -157,6 +157,7 @@ Set-Item -Path function:\PostPrompt  -Value $PostPrompt  -Options Constant
 
 [ScriptBlock]$Prompt = {
     $realLASTEXITCODE = $LASTEXITCODE
+    $host.UI.RawUI.WindowTitle = Split-Path $pwd.ProviderPath -Leaf
     PrePrompt | Microsoft.PowerShell.Utility\Write-Host -NoNewline
     CmderPrompt
     Microsoft.PowerShell.Utility\Write-Host "`nλ " -NoNewLine -ForegroundColor "DarkGray"
--- a/vendor/profile.ps1
+++ b/vendor/profile.ps1
@@ -65,11 +65,6 @@ function checkGit($Path) {
     }
 }
 
-# Load special features come from posh-git
-if ($gitStatus) {
-    Start-SshAgent -Quiet
-}
-
 # Move to the wanted location
 # This is either a env variable set by the user or the result of
 # cmder.exe setting this variable due to a commandline argument or a "cmder here"
--- a/vendor/profile.ps1
+++ b/vendor/profile.ps1
@@ -49,9 +49,14 @@ try {
 $gitLoaded = $false
 function Import-Git($Loaded){
     if($Loaded) { return }
-    if(-not (Get-Module -Name Posh-Git -ListAvailable) ) {
+    $GitModule = Get-Module -Name Posh-Git -ListAvailable
+    if($GitModule | select version | where version -le ([version]"0.6.1.20160330")){
+        Import-Module Posh-Git > $null
+    }
+    if(-not ($GitModule) ) {
         Write-Warning "Missing git support, install posh-git with 'Install-Module posh-git' and restart cmder."
     }
+    # Make sure we only run once by alawys returning true
     return $true
 }
 
--- a/vendor/profile.ps1
+++ b/vendor/profile.ps1
@@ -65,19 +65,6 @@ function checkGit($Path) {
     }
 }
 
-# Set up a Cmder prompt, adding the git prompt parts inside git repos
-function global:prompt {
-    $realLASTEXITCODE = $LASTEXITCODE
-    $Host.UI.RawUI.ForegroundColor = "White"
-    Write-Host $pwd.ProviderPath -NoNewLine -ForegroundColor Green
-    if($gitStatus){
-        checkGit($pwd.ProviderPath)
-    }
-    $global:LASTEXITCODE = $realLASTEXITCODE
-    Write-Host "`nλ" -NoNewLine -ForegroundColor "DarkGray"
-    return " "
-}
-
 # Load special features come from posh-git
 if ($gitStatus) {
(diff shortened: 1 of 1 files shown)


Jackbennett:
> Invoke-Expression doesn't add much overhead compared to the direct function call. I quickly smoke measured it with Measure-Command before PR and saw that it adds about 
0.001 to 0.003 secs compared to direct invocation of the function which measured, with an empty function, to ~0.008)

This is super interesting I'll experiment this. I haven't perf tested this PR yet it's welcome to changes.

janschulz:
There is also the idea to set the prompt in the clink lua code. at least for teh cmd path, this prevents the problem that you get a garbeled prompt when you call a new cmd:

```
c:\data\external\R\IRkernel (envs)
λ cmd
Microsoft Windows [Version 6.1.7601]
Copyright (c) 2009 Microsoft Corporation. Alle Rechte vorbehalten.

c:\data\external\R\IRkernel {git}{hg}
{lamb}
```

This problem really annoys me when I have conda build files which echo the commands they are using to build a package...

Not sure of that can be used for PS as well...

This is what I use (but the above is actually happening due to the changed prompt in init.bat -> That would have to be removed...):

```lua
---
 -- after conda activate: reset prompt 
---
function reset_prompt_filter()
    -- reset to original, e.g. after conda activate destroyed it...
    if string.match(clink.prompt.value, "{lamb}") == nil or not string.starts(clink.prompt.value,"\x1b[") then
        -- orig: $E[1;32;40m$P$S{git}{hg}$S$_$E[1;30;40m{lamb}$S$E[0m
        -- color codes: "\x1b[1;37;40m"
        cwd = clink.get_cwd()
        prompt = "\x1b[1;32;40m{cwd} {git}{hg} \n\x1b[1;30;40m{lamb} \x1b[0m"
        new_value = string.gsub(prompt, "{cwd}", cwd)
        clink.prompt.value = new_value
    end
end

clink.prompt.register_filter(reset_prompt_filter, 10)
```

centur:
Cool, I like your way more than my proposed hack. I haven't put much thinking into side effects and robustness, just sought some ways to preserve my prompt customization outside of profile.ps1. Happy to pass this feature initiative to you.

Side note - I played a bit with simple Stopwatch timing inside profile.ps1 and found that it's pretty slow, on my laptop it runs for almost 2 seconds - very noticeable for a fast and lightweight console. Removing some backward-compatibility bits and modules I shaved off more than a second from a console start time. It might be worth to do some testing and restructure it a bit - so an interested user can enable\disable certain features (e.g. posh-git is the beast) from his own profile and improve PowerShell start experience.
I wish someone who knows how these things are tied together would do the same for pre-posh scripts - because even with very lightweight configuration new console doesn't start immediately, start delay is very noticeable, and it's a painful experience on Windows.

Jackbennett:
Those `. .\imports` can get killed on the profile startup. They're only needed for W7 so we should really do some feature detection for that. Even so, PS still isn't an instant-on thing which I do find very annoying. But I pretty much never close PS.

My checkGit function in there should fend off posh-git as long as your startup dir isn't a git directory and needs it.

@janschulz clink isn't injected at all for powershell it doesn't need it. That was my point behind the failed move to `PS Default` because we keep getting really weird issues because of clink.

janschulz:
> clink isn't injected at all for powershell it doesn't need it. 

Ok, didn't know, I only use cmd :-)

Jackbennett:
I don't use cmd, it's nice to hear where we've had similar problems. Never sorry needed for jumping in.

centur:
@Jackbennett  disabling posh-git itself is a good step towards speeding up as it's slows down the profile boot time. I'd rather have some explicit way to load it (I have it anyway Import-Module posh-git) into a given console than to load it everywhere... Let's see how this PR goes through and then see how we can optimise other steps. 
Having 3-10 seconds startup time for a new console is not a great experience. It might be a powershell fault by itself, but it's extremely sad

Jackbennett:
Latest commits will require [this posh-git PR](https://github.com/dahlbyk/posh-git/pull/283) landing and a module update.

Just noticed Post-git now warn they don't support PS2.0 So this might be a bigger pain than we like. If PS2.0 is a big deal for us we'll have to have our own fork of Posh-Git.

Jackbennett:
PR has been merged in :) Just need a release to be wrapped up for whenever they do that.

Jackbennett:
Might as well tackle #528 while I'm at it. Brings PS more into line with cmd.

Jackbennett:
How do you poke CI to run again? That looks like some download fail. @MartiUK

janschulz:
Close and reopen or run `git commit --amend` followed by a `git push -f`

MartiUK:
Will we need to include something to tell users to update posh-git @Jackbennett? I foresee people having broken git statuses if they haven't.

MartiUK:
Okay to merge @Jackbennett ?

Jackbennett:
Yes I've tested what I can think off without any issues. Be nice for more people to have a look now.

Issue #749, opened 2015-12-01
Prevent other apps from overwriting the prompt?


janschulz:
Is there a way to prevent other programms from overwriting the prompt? I've `conda`, which adds the current environment to the start (and resets the cmder prompt back to the default) but also make the prompt behave very badly when I use 'up' to navigate to previous multiline commands (see screenshot: empty line between first and last and cursor in the empty line; I can edit, it just doesn't show up in the right place)

![2015-12-02_004433](https://cloud.githubusercontent.com/assets/890156/11517975/eaf32044-988d-11e5-9ac1-24d966efcee9.png)

vvartivarian-nbcu:
The multiline bug is in cmdr.lua get_git_status() caused by using os.execute(). I used io.popen instead here and it is fixed

danspam:
Suffering from the same double line bug here and using `io.popen()` instead of `os.execute()` does appear to cure it. Should be marked as a bug and fixed. Unfortunately I know nothing about lua so I would be completely unqualified to know if that is the correct fix and to issue a PR

JanSchulz:
I can also confirm that the extra-line bug is gone when using `io.popen()`. Will send a PR...

JanSchulz:
Ok, waiting for news on #736, as that would conflict with this PR...

JanSchulz:
I've a working solution which sets the prompt in my personal `config\cmder.lua` (after #736). Part of it sets the prompt in lua if the current prompt does not include a `{lamb}` anymore. An idea would be to remove the `prompt` call from `init.bat` and add such a filter to `vendor\cmder.lua`.

```
---
 -- Find out the basename of a file/directory (last element after \ or /
 -- @return {basename}
---
function basename(inputstr)
        sep = "\\/"
        local last = nil
        local t={} ; i=1
        for str in string.gmatch(inputstr, "([^"..sep.."]+)") do
                --t[i] = str
                --i = i + 1
                last = str
        end
        return last
end

---
 -- Find out current conda env
 -- @return {false|conda env name}
---
function get_conda_env()
    env_path = clink.get_env('CONDA_DEFAULT_ENV')
    if env_path then
        basen = basename(env_path)
        return basen
    end
    return false
end

---
 -- after conda activate: reset prompt and add conda env name 
---
function conda_prompt_filter()
    -- reset to original, e.g. after conda activate destroyed it...
    if string.match(clink.prompt.value, "{lamb}") == nil then
[...]

PR #961, opened 2016-05-24
cmd: change the prompt in lua

janschulz:
This keeps the PROMPT variable as is and changes the prompt to the cmder style
in the clink code.

This has two advantages:

* opening a cmd in a cmder session will now show the old prompt code instead of
  a ugly raw prompt without the replacements. This led to ugly output when a
  batch file echoed their content (e.g `conda build recipe/`).

* when a command rewrites the prompt (e.g. an activate in a virtualenv), these
  commands sometime simple overwrite the PROMPT so that the cmder enhancements
  were not anymore in place. Now we simply don't care and overwrite it with our
  stuff. This might mean that a user has to install a lua script so that e.g.
  conda environments are visible on the prompt.

Closes: https://github.com/cmderdev/cmder/issues/749

### DIFF
--- a/vendor/clink.lua
+++ b/vendor/clink.lua
@@ -12,6 +12,21 @@ dofile(clink_lua_file)
 
 -- now add our own things...
 
+---
+ -- Setting the prompt in clink means that commands which rewrite the prompt do
+ -- not destroy our own prompt. It also means that started cmds (or batch files
+ -- which echo) don't get the ugly '{lamb}' shown.
+---
+function set_prompt_filter()
+    -- orig: $E[1;32;40m$P$S{git}{hg}$S$_$E[1;30;40m{lamb}$S$E[0m
+    -- color codes: "\x1b[1;37;40m"
+    cwd = clink.get_cwd()
+    prompt = "\x1b[1;32;40m{cwd} {git}{hg} \n\x1b[1;30;40m{lamb} \x1b[0m"
+    new_value = string.gsub(prompt, "{cwd}", cwd)
+    clink.prompt.value = new_value
+end
+
+
 function lambda_prompt_filter()
     clink.prompt.value = string.gsub(clink.prompt.value, "{lamb}", "λ")
 end
@@ -230,6 +245,8 @@ function git_prompt_filter()
     return false
 end
 
+-- insert the set_prompt at the very beginning so that it runs first
+clink.prompt.register_filter(set_prompt_filter, 1)
 clink.prompt.register_filter(lambda_prompt_filter, 40)
 clink.prompt.register_filter(hg_prompt_filter, 50)
 clink.prompt.register_filter(git_prompt_filter, 50)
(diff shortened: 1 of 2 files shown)


janschulz:
Example old:

```
C:\Users\jschulz
λ cmd
Microsoft Windows [Version 6.1.7601]
Copyright (c) 2009 Microsoft Corporation. Alle Rechte vorbehalten.

C:\Users\jschulz {git}{hg}
{lamb} exit

C:\Users\jschulz
λ
```

new:

```
c:\data\external\cmder (prompt-in-clink)
λ cmd
Microsoft Windows [Version 6.1.7601]
Copyright (c) 2009 Microsoft Corporation. Alle Rechte vorbehalten.

c:\data\external\cmder>exit

c:\data\external\cmder (prompt-in-clink)
λ
```

janschulz:
A filter which adds a conda environment string in the front of the lambda:

```
---
 -- Find out the basename of a file/directory (last element after \ or /
 -- @return {basename}
---
function basename(inputstr)
        sep = "\\/"
        local last = nil
        local t={} ; i=1
        for str in string.gmatch(inputstr, "([^"..sep.."]+)") do
                --t[i] = str
                --i = i + 1
                last = str
        end
        return last
end

---
 -- Find out current conda env
 -- @return {false|conda env name}
---
function get_conda_env()
    env_path = clink.get_env('CONDA_DEFAULT_ENV')
    if env_path then
        basen = basename(env_path)
        return basen
    end
    return false
end

---
 -- add conda env name 
---
function conda_prompt_filter()
    -- add in conda env name
    local conda_env = get_conda_env()
    if conda_env then
        clink.prompt.value = string.gsub(clink.prompt.value, "{lamb}", "["..conda_env.."] {lamb}")
    end
end

clink.prompt.register_filter(conda_prompt_filter, 20)
```

Looks like this:

```
c:\data\external\cmder (prompt-in-clink)
λ activate borked35
Deactivating environment "C:\portabel\miniconda"...
Activating environment "C:\portabel\miniconda\envs\borked35"...

c:\data\external\cmder (prompt-in-clink)
[borked35] λ
```

janschulz:
Just for the record: I'm running my cmder with this PR since a few days and it works great: no more garbled output in `conda build` :-)

nullpunkt:
I would like this a +1 hence my installation actually keeps displaying {git}{lamda} and the other placeholders even though i already installed 0.4.7 of clink.

unbenannt

janschulz:
@nullpunkt I don't think this is the same "problem"  :-( This only "fixes" the problem that a subshell gets a proper commandline, not the one which is started in cmder.

-> please discuss in a different issue

janschulz:
rebased on top of current development

MartiUK:
:+1: LGTM

Stanzilla:
Merged. Thank you for that! :)

Issue #186, opened 2014-04-28
Add hotkey to switch tabs


lnguyen639:
Please add the hotkeys Ctrl+PgUp/Ctrl+PgDn to move between tabs in the console. 
More hotkeys if possible:
.) Ctrl+Shift+PgUp/PgDn -> move tab's position
.) Hot key to open Cygwin console (or console of choice)
.) Hot key to rename tab

Thank you!

MartiUK:
I've move the milestone to 1.3 for now.

DoCode:
+1

Morten242:
Moving the active tab is already in there: Win+Alt+Left/Right (can be remapped in Keys&Macros)

And CTRL+Tab functionality can be set to move to adjecent tabs by unchecking "Recent mode" in Main->Tabs in settings.

Next and Previous tab can also be remapped.

rgeminas:
To map Ctrl + PageUp / Ctrl + PageDown to move tabs:

Under macros:

Ctrl + PageUp => Tab(4, -1); Tab(1);
Ctrl + PageDown => Tab(4, 1); Tab(1);

Jackbennett:
Pending PR Decision then.

lucatrv:
Adding CTRL-PgUp / CTRL-PgDown for tab switching is super convenient, I wonder why this has not been added to the default configuration yet.

slugmandrew:
Thanks for the help on here!

kristianperkins:
CTRL+Tab and CTRL+Shift+Tab are mapped for tab switching

PR #963, opened 2016-05-25
Disable history switching behavior of ctrl+tab. Sequential switching.

Jackbennett:
With regards to #186 pull this in if we still want to change this default.

Do note: If you ctrl+tab once, holding control you can still use the arrow keys.


Henry-T:
Why not make it customizable? I want Alt+J Alt+K to move around.

Jackbennett:
@Henry-T It's always customizable, this patch is just the defaults.

Issue #809, opened 2016-01-20
No longer shows git current branch


jmiazga:
Cmder v1.2.9 no longer shows the current branch in git. Instead it shows the name and version from package.json.

![image](https://cloud.githubusercontent.com/assets/181441/12459594/9bf65350-bf7d-11e5-843a-1fd70c944c3e.png)

MartiUK:
Calling @vladimir-kotikov

vladimir-kotikov:
@jmiazga, does this happen only in this particular directory, or in every git repo?

Could you please paste the output of `git rev-parse --git-dir --is-inside-git-dir --is-bare-repository --is-inside-work-tree` (@MartiUK, this is still a nice way to debug :smile: )

jmiazga:
It happens in every git repo. If there is no package.json nothing is displayed. I downloaded v1.2 and everything works as expected.

Output:
.git
false
false
true

vladimir-kotikov:
Hm, that's strange. Just checked and the content of `config/cmder.lua` from 1.2.9 is strictly equal to the one from 1.2.0. So it seems the only reason is the different git versions.

keathmilligan:
This seems to be a Windows environment issue. The redirect in the io.popen statement in the get_git_branch function in cmd.lua is causing the lines() function to return nothing in some cases.

One computer (that doesn't have the issue):

```
Lua 5.3.2  Copyright (C) 1994-2015 Lua.org, PUC-Rio
> for line in io.popen("git branch 2>nul"):lines() do print(line) end
* master
>
```

Another computer running the same LUA and git:

```
Lua 5.3.2  Copyright (C) 1994-2015 Lua.org, PUC-Rio
> for line in io.popen("git branch 2>nul"):lines() do print(line) end
>
```

You can remove the "2>nul" as work-around for now.

PR #971, opened 2016-05-31
Fix the git command to obtain the dirty state of the repository

### LINKED ISSUE #928: git (dirty) repository status not shown in promtp since 1.3.0pre
I'm not getting the indicator of a dirty repo in the prompt since I updated to 1.3.0pre.

The command I think it's getting executed is `git diff --quiet --ignore-submodules HEAD 2>nul` but in my current instalattion returns:

```
fatal: ambiguous argument '2': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

I'll try to obtain a command that works and commit it as a fix later.

sveggiani:
Fixes #928

### DIFF
--- a/vendor/clink.lua
+++ b/vendor/clink.lua
@@ -5,7 +5,7 @@
 
 
 -- At first, load the original clink.lua file
--- this is needed as we set the script path to this dir and therefore the original 
+-- this is needed as we set the script path to this dir and therefore the original
 -- clink.lua is not loaded.
 local clink_lua_file = clink.get_env('CMDER_ROOT')..'\\vendor\\clink\\clink.lua'
 dofile(clink_lua_file)
@@ -193,7 +193,7 @@ end
  -- @return {bool}
 ---
 function get_git_status()
-    return io.popen("git diff --quiet --ignore-submodules HEAD 2>nul")
+    return os.execute("git diff --quiet --ignore-submodules HEAD")
 end
 
 function git_prompt_filter()
@@ -240,4 +240,3 @@ for _,lua_module in ipairs(clink.find_files(completions_dir..'*.lua')) do
         dofile(filename)
     end
 end
-


luisrudge:
I can confirm this is now working:
image

Stanzilla:
We had a few problems with os.execute which is why we switched away from it https://github.com/cmderdev/cmder/issues/749

Does this help you maybe? https://github.com/cmderdev/cmder/issues/809#issuecomment-207203013

sveggiani:
@Stanzilla nope, it doesn't work. I'll try to come up with another solution.

vBm:
@Stanzilla I've also tried what's suggested in 809 but doesn't help either.
However, this patch makes it work again.

Stanzilla:
https://github.com/cmderdev/cmder/commit/567889e69fb00d7a3f7f661b8552916cb7e03a3d found it, paging @janschulz

janschulz:
Unfortunately, I only send the patch, not found the cause: @vvartivarian-nbcu suggested the particular fix: https://github.com/cmderdev/cmder/issues/749#issuecomment-161407632

janschulz:
But reading this: http://www.lua.org/manual/5.1/manual.html#pdf-io.popen it looks like the solution could never work: t returns a handle and not the exit code:

> io.popen (prog [, mode])
> Starts program prog in a separated process and returns a file handle that you can use to read data from this program (if mode is "r", the default) or to write data to this program (if mode is "w").
> This function is system dependent and is not available on all platforms.

janschulz:
This might work: http://stackoverflow.com/a/14350308/1380673

janschulz:
See also http://stackoverflow.com/questions/7607384/getting-return-status-and-program-output

this works:
```lua
---
 -- Get the status of working dir
 -- @return {bool}
---
function get_git_status()
    local file = io.popen("git diff --quiet --ignore-submodules HEAD 2>nul")
    -- This will get a table with some return stuff
    -- rc[3] will be the signal, 0 or 1
    local rc = {file:close()}
    return rc[3] == 0 
end
```

-> Will send a PR

janschulz:
See https://github.com/cmderdev/cmder/pull/974 for the PR, please test... :-)

sveggiani:
Closing this in favor of the more appropiate fix in #974

PR #974, opened 2016-06-02
Fix git branch name never shown as dirty

janschulz:
The problem was that io.popen() returns a file handle and not the return code of the
called program. The problem was introduced by 567889e.

The new code was inspired by
http://stackoverflow.com/a/14031974/1380673

Closes: https://github.com/cmderdev/cmder/pull/971
Closes: https://github.com/cmderdev/cmder/issues/928

### DIFF
--- a/vendor/clink.lua
+++ b/vendor/clink.lua
@@ -193,7 +193,11 @@ end
  -- @return {bool}
 ---
 function get_git_status()
-    return io.popen("git diff --quiet --ignore-submodules HEAD 2>nul")
+    local file = io.popen("git diff --quiet --ignore-submodules HEAD 2>nul")
+    -- This will get a table with some return stuff
+    -- rc[3] will be the signal, 0 or 1
+    local rc = {file:close()}
+    return rc[3] == 0
 end
 
 function git_prompt_filter()


janschulz:
A different way would be to use the hg wy (read lines and see if there are any) with `git status -s -uno`.

sveggiani:
Hi @janschulz.
I've done some test and this works. However, if I run the `git diff --quiet --ignore-submodules HEAD 2` commnad with **git version 2.7.4.windows.1** the output is always:

```
"fatal: ambiguous argument '2': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'"
```

I'm a little curious about why it works. Do you know why? 

Thanks!

janschulz:
@sveggiani the `2` is not part of the git command, but part of the redirection: in this case it means that the file handle `2` (`stderr`, which gets the error output) should be redirected to the special file NUL (aka "into nowhere"). In effect this means that error output should be hidden.

See here fore more info: https://www.microsoft.com/resources/documentation/windows/xp/all/proddocs/en-us/redirection.mspx?mfr=true

sveggiani:
@janschulz ohh, I see. Thank you very much ;)

PR #999, opened 2016-06-15
Suppress profile load error

ppejovic:
Suppress profile error when there are no ps1 files to source in config\profile.d


ppejovic:
`ls` was aliased to `ls.exe` in my shell. This is not a valid bug fix.

PR #1042, opened 2016-07-17
Added kill ssh-agent to build.ps1

daxgames:
Kills ssh-agent.exe if it is running from the $env:cmder_root we are building.


daxgames:
Did I open that or to go into master?  I do I meant to open it against development.

Stanzilla:
That went into master and I'm actually fine with that, not sure if I want to keep using the dev branch

Issue #1057, opened 2016-08-03
Python virtualenv not activating in Windows 10 Cmder


cwmat:
I can initialize a virtualenv environment using Cmder but when I run the following command:

`env/Scripts/activate`

Nothing happens (i.e. "(env)" doesn't show up in front of my path in the terminal as its supposed to).

When running the same in regualr cmd it works as intended.  

Any thoughts?  I was thinking maybe there is some Windows setting that might be preventing Cmder from running a batch file (activate.bat) or a file without an extension (activate).  I ran Cmder as admin and it did not change anything though.  I posted the full question and background here: http://superuser.com/questions/1108063/python-virtualenv-not-activating-in-windows-10-cmder

janschulz:
See here: https://github.com/cmderdev/cmder/issues/1056

I think we should do that automatically...

cwmat:
Thanks @janschulz .  I did the following (below) but I am still running into the same problem.  Just want to confirm that the following procedure was correct:

I made a file called "conda.lua" in the config folder with the contents as suggested in #1056 

I then added the following line to /vendor/clink.lua:

`-- now add our own things...`
`local conda_lua_file = clink.get_env('CMDER_ROOT')..'\\config\\conda.lua'`
`dofile(conda_lua_file)`

Was this the correct approach?  Any ideas why I still might be having issues getting conda to display?

Thank you.

janschulz:
Oups, sorry... No, don't change anything in vendor, only place a file in config. The conda example needs to be adjusted (conda is an alternative to virtualenv) . In some place it looks for a environment variable `CONDA_DEFAULT_ENV` and this needs to be replaced with whatever virtualenv uses for this purpose.

cwmat:
Gotcha!  Sorry for the confusion.  That worked.  Thank you.  Just changed the `CONDA_DEFAULT_ENV` to `VIRTUAL_ENV `and it worked.  

It did throw this though:
`C:\PROGRA~1\CMDER_~1\config\venv.lua:8: attempt to call global 'basename' (a nil value)`

I don't know much about LUA.  Any reason you can think of off the top of your head as to why this message would get shown?  If not no worries!  You have given me enough information for me to go do some research and I can close the issue.  

Thank you for all of the help so far!

lpinner:
@cwmat see https://github.com/cmderdev/cmder/issues/1056#issuecomment-237403714 and http://superuser.com/a/1108564/195565

lpinner:
[...]

Issue #1054, opened 2016-07-31
prompt is Garbled


zydxhs:
As shown below：
![prompt](https://cloud.githubusercontent.com/assets/15759657/17278173/ca3ff79a-5788-11e6-82e9-09d44a364954.png)

cmder ver is 1.3 in win7 64bit。

janschulz:
Seems that the replacement code can't handle non-ascii chars?

```
c:\temp
λ mkdir ü

c:\temp
λ cd ü

c:\temp\�
λ
```

janschulz:
We [currently](https://github.com/cmderdev/cmder/blob/master/vendor/clink.lua#L23-L26) do this:

```lua
    cwd = clink.get_cwd()
    prompt = "\x1b[1;32;40m{cwd} {git}{hg} \n\x1b[1;30;40m{lamb} \x1b[0m"
    new_value = string.gsub(prompt, "{cwd}", cwd)
    clink.prompt.value = string.gsub(new_value, "{lamb}", "λ")
```

So my reading is that we can't pass the output of `clink.get_cwd()` to `clink.prompt.value`?

CC: @mridgers (clink maintainer)

zydxhs:
May be the clink does not support non-ascii.

janschulz:
I think it does (letting the cmd set the prompt and then handling it in clink works...), just that the string has to get encoded... No idea how that works :-/

janschulz:
So, a bit of debugging (to a ANSI encoded file) shows this:

```
cwd(string): c:\temp\ü
prompt(string): c:\temp\Ã¼> # before any changes in lua
prompt(string): [1;32;40mc:\temp\ü {git}{hg}  # after changes in lua
```

If I interpete http://stackoverflow.com/a/14981670/1380673 right, then clink needs to translate the prompt between utf (what clink/lua seems to use in `get_cwd()`) and whatever is done with `clink.prompt.value` when it is written back to the prompt.

[Not, I have changed my codepage to utf8: `chcp  65001` but the same debug output happens with `chcp 850` which is the default in a regular cmd window]

janschulz:
I've taken that upstream: https://github.com/mridgers/clink/issues/415

zydxhs:
Directly set value to "clink.prompt.value" is ok:

janschulz:
I can't reproduce: on my system with `clink.prompt.value = clink.get_cwd()` (not the last `()`!) exactly the same happens:

```
c:\temp\�
```

[...]

Issue #1056, opened 2016-08-02
conda environment not displaying


cmorgan:
the conda env, usually looks like `(dev) c:\dev\code (master)`
is missing from version 1.30 full distribution

janschulz:
As we currently do not use the original prompt, you have to add conda information yourself: 

Add a file called `<cmder dir>\config\<somehing>.lua` with the following content:

```lua
---
 -- Find out current conda env
 -- @return {false|conda env name}
---
function get_conda_env()
    env_path = clink.get_env('CONDA_DEFAULT_ENV')
    if env_path then
        basen = basename(env_path)
        return basen
    end
    return false
end

---
 -- add conda env name 
---
function conda_prompt_filter()
    -- add in conda env name
    local conda_env = get_conda_env()
    if conda_env then
        clink.prompt.value = string.gsub(clink.prompt.value, "{lamb}", "["..conda_env.."] {lamb}")
    end
end

clink.prompt.register_filter(conda_prompt_filter, 20)
```

lpinner:
At the point this runs, the "{lamb}" has already been replaced with "λ".

To get it working, I replaced:
`clink.prompt.value = string.gsub(clink.prompt.value, "{lamb}", "["..conda_env.."] {lamb}")`

with:
`clink.prompt.value = string.gsub(clink.prompt.value, "λ", "["..conda_env.."] λ")`

janschulz:
Please try https://github.com/cmderdev/cmder/pull/1070 for a version which parses the original prompt.

PR #1070, opened 2016-08-09
Parse the original prompt for cwd and env names

janschulz:
clink.get_cwd() is returning a string which is differently encoded than what
clink.prompt.value expects. This results in garbled path names if the path
condains non-ASCII chars. The (arguable hacky) solution is to parse the old
prompt for the current directory (which breaks if the user sets a PROMPT env var
which is incompatible to the regex used here...).

Also parse out a environment name set by systems like virtualenv or conda: this
could be done more specifically by targeting each such system and using the
usually set environment variable but this would mean that we would have to do
that for each and every such system out there and that is probably not a sane
idea...

Closes: https://github.com/cmderdev/cmder/issues/1054 (garbled prompt)
Closes: https://github.com/cmderdev/cmder/issues/1057 https://github.com/cmderdev/cmder/issues/1056 (environment issues)

Example session:

```cmd
C:\Users\jschulz
λ activate py35

C:\Users\jschulz
(py35) λ cd c:\temp\ü\(hallo)\

c:\temp\ü\(hallo)
(py35) λ
```

A "better" solution would be proper unicode support in clink: https://github.com/mridgers/clink/issues/415

### DIFF
--- a/vendor/clink.lua
+++ b/vendor/clink.lua
@@ -18,12 +18,34 @@ dofile(clink_lua_file)
  -- which echo) don't get the ugly '{lamb}' shown.
 ---
 function set_prompt_filter()
+    -- get_cwd() is differently encoded than the clink.prompt.value, so everything other than
+    -- pure ASCII will get garbled. So try to parse the current directory from the original prompt
+    -- and only if that doesn't work, use get_cwd() directly.
+    -- The matching relies on the default prompt which ends in X:\PATH\PATH>
+    -- (no network path possible here!)
+    local old_prompt = clink.prompt.value
+    local cwd = old_prompt:match('.*(.:[^>]*)>')
+    if cwd == nil then cwd = clink.get_cwd() end
+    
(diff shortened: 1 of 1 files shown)


jing2si:
nice work~

codesboy:
@janschulz @jing2si 
Good job! Buddy. ⊙o⊙

Stanzilla:
@janschulz do you still want to wait for a response to your clink issue or won't that change what is in this PR?

janschulz:
@Stanzilla IMO this is a nice shortterm fix for the problem. If there is a better way (=proper unicode support in clink) , I will add a new PR.

davpaez:
According to [this](https://github.com/chrisant996/clink/commit/99b2b6566ab59a691ea4b501c4ea0adfedb14f17) commit in clink, this issue appears to have been fixed.

If so, maybe the `clink.lua` script should use the `clink.get_cwd()` inside the `set_prompt_filter()` function again, instead of relying on the more convoluted `clink.prompt.value`?

PR #1078, opened 2016-08-20
Fix bash login when $CMDER_ROOT has spaces

orionlee:
The patch fixed the errors when using bash and cmder is installed under a directory with spaces in the path, e.g. `c:\Users\Foo Bar\cmder` . Please review. Thanks.

The first commit fixed errors I encountered (in v1.3.0), while the second commit fixed similar issues.

Note: The development branch (the base) is behind master, in fact, pre v1.3.0.  Should I stick with merging to it?

### DIFF
--- a/vendor/cmder.sh
+++ b/vendor/cmder.sh
@@ -43,7 +43,7 @@ fi
 
 if [ -d "${CMDER_ROOT}/config/profile.d" ] ; then
   unset profile_d_scripts
-  pushd ${CMDER_ROOT}/config/profile.d >/dev/null
+  pushd "${CMDER_ROOT}/config/profile.d" >/dev/null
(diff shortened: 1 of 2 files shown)


orionlee:
FYI, the original errors I encountered:

```
bash: pushd: /c/Users/Foo: No such file or directory
bash: [: /c/Users/Foo: binary operator expected
bash: /c/Users/Foo: No such file or directory
```

Stanzilla:
No, please change to master. @daxgames are you ok with this? I mean it's sort of a trivial change (at least the first commit) but you wrote the code in the first place.

Thanks for the PR, orionlee!

orionlee:
@Stanzilla I agree 2nd commit is slightly tricky (double quotes cannot handle having spaces along with a wild card),  and is more prone to regress.

daxgames:
Looks good to me.

Stanzilla:
Thank you!

PR #1101, opened 2016-09-27
Added closing process in get_git_status

alexandr-san4ez:
After execution command: git add * git continues to operate without closing. The process does not stop.

### DIFF
--- a/vendor/clink.lua
+++ b/vendor/clink.lua
@@ -203,9 +203,12 @@ end
  -- @return {bool}
 ---
 function get_git_status()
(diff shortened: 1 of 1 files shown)


KcPele:
fatal: Unable to create 'C:/Users/userpc/Desktop/Web app django/.git/index.lock': File exists.

If no other git process is currently running, this probably means a
git process crashed in this repository earlier. Make sure no other git
process is running and remove the file manually to continue.

Please how do i solve these issue

PR #1131, opened 2016-10-14
Fix for slow startup under certain conditions (#1122)

### LINKED ISSUE #1122: Slow on startup
On my machine, a vanilla installation of Cmder takes >5s to startup compared to the instantaneous ConEmu. 
I tried disabling everything in the `init.bat` script but it didn't really make a difference.

lamarqua:
Fixes #1122 by simply calling CreateProcess before the SendMessageTimeout
calls that are used to propagate environment variable changes.


lamarqua:
I'm flying tomorrow and will not have access to a windows PC for two weeks - so here is the quick and dirty PR.

PR #1157, opened 2016-11-10
Fix typo in init.bat

winks:


janschulz:
Oups, thank for fixing, @winks!

PR #1189, opened 2016-12-06
Documented some undocumented tab related shorcuts

knaos:
Added documentation for tab switching shortcuts, which I find very useful.


Stanzilla:
thanks!

PR #1246, opened 2017-02-02
Update ConEmu to 170118 (stable)

ParallaxVT:
ConEmu has a new version available


Stanzilla:
This is currently labeled as a preview release, I only bump to stables, sorry

PR #1252, opened 2017-02-09
Fix #1245: Allow alias value to contain '=' characters

### LINKED ISSUE #1245: Create alias with equal sign gets truncated in cmder cli
Hey there as per title, when I try in cmder prompt to create an alias with an equal sign, what comes with the `=` gets truncated and not correctly stored in the `user-aliases.cmd` file.

For ex:
`λ alias wifi=netsh wlan show profile $1 key=clear`
will be stored as:
`wifi=netsh wlan show profile $1 ke`

I guess for `$reasons` cmder reads `y=clear` as `$something` worthy evaluating and so it's cut off from the actual alias storing process. Manually editing the file quickly overcome this issue.

Am I blatantly missing something on the ever growing *Effin' Manual* that I haven't accurately read? 😛

glucas:


Stanzilla:
thanks!

PR #1280, opened 2017-02-28
Allow profile.d ps1 file to override prompt

bdukes:
The *.ps1 files in the profile.d/ directory should be run _after_ the default prompt variables ($PrePrompt, $PostPrompt, and $CmderPrompt) are defined, so that those profile files can have access to overwrite them, otherwise the prompt can _only_ be set via user-profile.ps1.

### DIFF
--- a/vendor/profile.ps1
+++ b/vendor/profile.ps1
@@ -86,19 +86,6 @@ if (Get-Module PSReadline -ErrorAction "SilentlyContinue") {
 # Enhance Path
 $env:Path = "$Env:CMDER_ROOT\bin;$env:Path;$Env:CMDER_ROOT"
 
-# Drop *.ps1 files into "$ENV:CMDER_ROOT\config\profile.d"
-# to source them at startup.
-if (-not (test-path "$ENV:CMDER_ROOT\config\profile.d")) {
-  mkdir "$ENV:CMDER_ROOT\config\profile.d"
-}
-
-pushd $ENV:CMDER_ROOT\config\profile.d
-foreach ($x in ls *.ps1) {
-  # write-host write-host Sourcing $x
-  . $x
-}
-popd
-
 #
 # Prompt Section
 #   Users should modify their user-profile.ps1 as it will be safe from updates.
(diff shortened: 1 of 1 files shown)


Jackbennett:
`PrePrompt`, `CmderPrompt` and `PostPrompt` where intended to be small hooks to easily modify the prompt. I guess if you know what you're doing we do indeed need a method to replace the whole prompt. 

This needs further testing but I suspect if you move what you've already moved further down to line 178 you could leave #1282 out and simply allow something in `profile.d/` to overwrite `$prompt` just before it's set to read-only.  Might need to specify the prompts to global: scopes but the dot-source should pull those files into this scope. Thus as long as anything in `profile.d` sets a `prompt` variable it'll be used instead.

paravz:
hijacking this in regards to prompt customizing: I could not find a way to remove newline from the prompt using the existing hooks PrePrompt, CmderPrompt and PostPrompt, since new line is added in global $Prompt in vendor\prompt.ps1

I can easily customize the 'cmd' prompt via lua "filter", looking for direction for 'powershell' to create a proper pull

bdukes:
@paravz, #1282 is an attempt to allow removing the newline, though @Jackbennett had indicated that we might combine the approach here with that one.  But, yes, at the moment there doesn't appear to be a supported way to totally override the prompt.

Jackbennett:
Yeh I've tested what I said. If you just move the same lines you've already moved in this further down to line 178 after `[scriptblock]$prompt` before `set-item -Path function:\prompt...`

Any file in `profile.d` that exports `$prompt` will overwrite the cmder one. It must be a scriptblock type though or you'll get an error. Might be worth testing the type and writing a nicer error although the powershell one is pretty clear;
```
Cannot convert the "test" value of type "System.String" to type "System.Management.Automation.ScriptBlock".
At C:\f\src\cmder\config\profile.d\prompt.ps1:1 char:1
+ $prompt =  'test'
```

If you force-push this PR to update it I'll just merge it in. Otherwise I don't know my git-fu to do it myself but keep your credit.

bdukes:
@Jackbennett, I've pushed a change.  First, the default variables `$PrePrompt`, `$PostPrompt`, `$CmderPrompt`, and `$Prompt` are defined, then both `profile.d/*.ps1` scripts and the `user-profile.ps1` script are called, and finally the prompt variables are set at their `Function:\` location.

Let me know if there's anything else that needs to change.

Thanks!

Jackbennett:
I'll probably mess something up by not merging into development but we'll see.

Thanks you. There's some good room for a documentation and examples update if anyone watching add a PR.

PR #1282, opened 2017-03-01
Allow totally overriding Prompt

bdukes:
This allows using a system like oh-my-posh, without forcing the λ in the prompt


Jackbennett:
How are you assigning $prompt in the first place to make that `if` bypass the scriptblock prompt?

bdukes:
Right, a script in either `profile.d/` or `user-profile.ps1` would be able to set `$Prompt` (even without #1280 you could use `user-profile.ps1`, like you would do now for any other prompt customizations)

Jackbennett:
yes discuss on #1280

PR #1299, opened 2017-03-09
Changed the way directory is defined .git

alexandr-san4ez:
`clink.find_dirs` - do not specify the directory if there is Cyrillic in the way.
`clink.is_dir` - more logical and faster.

default


alexandr-san4ez:
`d:\sandbox\рус` - "**рус**" these are Cyrillic symbols.

Stanzilla:
Thanks!

PR #1310, opened 2017-03-17
Print command index in history

schaazzz:
Currently the "history" command in Cmder prints commands without any index associated with any command, so you can copy/paste a command or use "!!" to execute a command from the history but, unlike BASH, you can't do "!<command_index>".

"cat" has a switch, "-n", which prints line numbers of the file being printed. This, in conjunction with "history_io=3" in Clink settings (to read/write history when editing a command") works just like BASH, i.e., you can use the index printed next to a command in the history to execute that command as "!<command_index>".

Demo of how it works/looks:

image


schaazzz:
The change in __init.bat__ would take care of setting __history_io=3__ for the initial run (using the *__clink.history_io__* environment variable), if there is no settings file available but the settings file would still contain __history_io=0__ on subsequent runs.

If this pull request is accepted, a more permanent solution is needed for having __history_io__ set to __3__ in the generated settings file.

DRSDavidSoft:
@Stanzilla this seems so useful to have! Could you approve it?

Stanzilla:
It needs a solution for the problem mentioned in the comment above

daxgames:
@schaazzz You there?  This needs to be synced with Cmder Master and needs these changes in the following files so this can be merged:

* We do not touch users `config\settings` files after the first run so no need to set that in init.bat.  The following will set `history_expand_mode = 3` on the first run.  Users can set it very easly if the want to by typing `clink set history_expand_mode 3`  if they do not have the setting already.
* Don't need the `set clink.history_io=3` that you added in `vendor\init.bat`

https://github.com/cmderdev/cmder/blob/f39b921121f3c4c99741cc33ef6bb554f5b2a69d/vendor/clink_settings.default#L90

* Need the quotes
https://github.com/cmderdev/cmder/blob/f39b921121f3c4c99741cc33ef6bb554f5b2a69d/vendor/user_aliases.cmd.default#L12

Make these changes and we will merge.

daxgames:
See #2068 for completion of this feature.  Closing due to lack of response from OP.

PR #1360, opened 2017-04-28
Fixed quotation marks BUG

zthxxx:
1. Changed " --login -i&quot;" to "&quot; --login -i" in ConEmu
2. Closed quotation in "alias.bat :p_del doskey" command of "%ALIASES%"

Suppose %CMDER_ROOT% is "C:\Program Files\cmder",
this BUG will be raised while we exec command of

"unalias xxx"
or creat terminal "{bash::bash}"
and "{bash::mintty}",

it will throws an error like:
'C:\Program' is not recognized as an internal or external command,
operable program or batch file.


zthxxx:
@MartiUK I have tested it, and make sure that this will contribute to fix aliasing, then commit the PR.

PR #1372, opened 2017-05-12
#420 and #577 - use registry for CMDER_START

superelitist:
Environment variables don't seem to work properly when cmder starts
ConEmu using /single. These changes communicate the intended start path
using a registry key.

Sorry, I didn't make changes from development, but that branch looks pretty out of date.

### DIFF
--- a/launcher/src/CmderLauncher.cpp
+++ b/launcher/src/CmderLauncher.cpp
@@ -180,15 +180,16 @@ void StartCmder(std::wstring path, bool is_single_mode)
 		swprintf_s(args, L"/Icon \"%s\" /Title Cmder", icoPath);
 	}
 
-	SetEnvironmentVariable(L"CMDER_ROOT", exeDir);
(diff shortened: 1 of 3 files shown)


DRSDavidSoft:
@Stanzilla on the subject of branches, if the `development` isn't getting much activity and even seems to be obsolete, could you remove it? I was also confused when I first tried to make a PR.

Stanzilla:
That would certainly be a an option. @cmderdev/trusted-contributors thoughts?

daxgames:
No issues.

superelitist:
Sorry it took me so long. I'd never rebased anything before. I think I did it correctly.

Stanzilla:
Not quite, but almost, you commited merge conflicts here https://github.com/cmderdev/cmder/pull/1372/files#diff-159626d08a0475178d80ae25e9d36f17R91 and in init.bat

Issue #193, opened 2014-05-14
How to enable ssh-agent?


CaseyLabs:
I'm trying to setup Cmder to use ssh-agent forwarding:

λ ssh-agent
SSH_AUTH_SOCK=/tmp/ssh-JqBCMa6092/agent.6092; export SSH_AUTH_SOCK;
SSH_AGENT_PID=7840; export SSH_AGENT_PID;
echo Agent pid 7840;

λ ssh-add my-public-key.pub
Could not open a connection to your authentication agent

Any ideas what I'm doing wrong here?

Thank!

cdelgehier:
It is working for me but the agent forwarding is not good
How to export SOCK and  PID variable?

cdelgehier:
With your script @teleivo and http://niki.hammler.net/wiki/PuttyAgent_now_supports_openssh/cygwin/msys 
I am a satisfied man

IrNoThnk:
has anyone successfully had confirmation with ssh-add working?

ssh-add -c /path/to/key

Adding is no problem, but you'll get a failure when connecting because there is no /usr/libexec/ssh-askpass. At least that's the issue that I've had in Ubuntu before.

isdsc:
I am currently using `start-ssh-agent` script [provided by msysgit](https://github.com/msysgit/msysgit/blob/master/cmd/start-ssh-agent.cmd) to set up my ssh-agent in cmder (it is in the default cmder search path: %git_install_root%\cmd\). Is there any reason to switch to the one added to the [development branch](https://github.com/bliker/cmder/blob/development/bin/agent.cmd)?

markwint:
@isdsc any idea how to make it enable agent forwarding?

isdsc:
@markwint if the question is how `start-ssh-agent` works with agent forwarding, the answer is (AFAIK):
> It just starts `ssh-agent` properly in your session [in Windows console] and adds the default identity, it doesn't deal with agent forwarding.

Here is my ssh configuration:

```
λ head ~/.ssh/config
# Shortcut to connect to my server
Host mysvc
 User MyName
 IdentityFile ~/.ssh/id_rsa
 HostName myservice.example.com

# GitHub account: isdsc
Host github.com
 IdentityFile ~/.ssh/isdsc
 HostName github.com
```

Here is how I use agent forwarding with a second identity to connect to GitHub. See [GitHub](https://developer.github.com/guides/using-ssh-agent-forwarding/) for details.

```
λ start-ssh-agent
[...]

PR #1391, opened 2017-06-05
Added pagent ssh auth support

AlphaGit:
(and some comments on the user-profile.cmd file)

Inspired by the comments from #193 and my personal need to use pageant instead of OpenSSH authentication agents (which is more Window user-friendly), I have used this approach which works as expected.

Keeping the spirit of the current scripts, I left it disabled, and with some comments explaining what they all do.


Stanzilla:
Thanks

AlphaGit:
@Stanzilla You're very welcome!

PR #1399, opened 2017-06-08
Fix Cmder not changing to CMDER_START

Hugo-KTM:
Fix to allow Cmder to change to different startup directory (drive root) if already on the same drive as CMDER_START
possibly relevant: HOME != USERPROFILE

### DIFF
--- a/vendor/init.bat
+++ b/vendor/init.bat
@@ -160,7 +160,11 @@ if not defined HOME set "HOME=%USERPROFILE%"
 :: This is either a env variable set by the user or the result of
 :: cmder.exe setting this variable due to a commandline argument or a "cmder here"
 if defined CMDER_START (
-    pushd "%CMDER_START%"
+    if "%CMDER_START:~-1%" neq "\" (
+        SET CMDER_START="%CMDER_START%\"
+    )
+
+    cd /d "!CMDER_START!"
 )
 
 if exist "%CMDER_ROOT%\config\user-profile.cmd" (
--- a/vendor/init.bat
+++ b/vendor/init.bat
@@ -159,13 +159,15 @@ if not defined HOME set "HOME=%USERPROFILE%"
 
 :: This is either a env variable set by the user or the result of
 :: cmder.exe setting this variable due to a commandline argument or a "cmder here"
+setlocal enabledelayedexpansion
 if defined CMDER_START (
     if "%CMDER_START:~-1%" neq "\" (
-        SET CMDER_START="%CMDER_START%\"
+        SET "CMDER_START=%CMDER_START%\"
     )
 
-    cd /d "!CMDER_START!"
+    pushd "!CMDER_START!"
 )
+endlocal
 
 if exist "%CMDER_ROOT%\config\user-profile.cmd" (
     REM Create this file and place your own command in there
--- a/vendor/init.bat
+++ b/vendor/init.bat
@@ -160,7 +160,7 @@ if not defined HOME set "HOME=%USERPROFILE%"
 :: This is either a env variable set by the user or the result of
 :: cmder.exe setting this variable due to a commandline argument or a "cmder here"
 if defined CMDER_START (
-    cd /d "%CMDER_START%"
+    cd /d "%CMDER_START%\"
 )
 
 
--- a/vendor/init.bat
+++ b/vendor/init.bat
@@ -160,9 +160,11 @@ if not defined HOME set "HOME=%USERPROFILE%"
 :: This is either a env variable set by the user or the result of
 :: cmder.exe setting this variable due to a commandline argument or a "cmder here"
 if defined CMDER_START (
-    cd /d "%CMDER_START%\"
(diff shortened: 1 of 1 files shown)


Stanzilla:
Can you explain this a bit more?

Hugo-KTM:
Environment
===========
```cmd
USERPROFILE=C:\Users\<User>
HOME=H:\
REM set by Cmder:
CMDER_ROOT=C:\Users\<User>\bin\PortableApps\cmder
```

Registry
========
```reg
[... Classes\Drive\shell\Cmder\command]
@="\"C:\\Users\\<User>\\bin\\PortableApps\\cmder\\Cmder.exe\" \"%V\""

[... Classes\Directory\shell\Cmder\command]
@="\"C:\\Users\\<User>\\bin\\PortableApps\\cmder\\Cmder.exe\" \"%V\""

[... Classes\Directory\Background\shell\Cmder\command]
@="\"C:\\Users\\<User>\\bin\\PortableApps\\cmder\\Cmder.exe\" \"%V\""
```

Steps
=====
1. run Cmder  
   Cmder starts at %USERPROFILE%
2. start Cmder via conext menu on C:\ (Drive or Directory Background)  
   Cmder starts at %USERPROFILE%

My edit changes the behaviour at step 2 so that Cmder starts at C:\ instead of %USERPROFILE%
When starting Cmder at a specific location I expect Cmder to start there. That is what I tried to fix.
The Problem only occurs when (at step 2) you try to run Cmder at the drive root.

vendor/init.bat
```diff
@@ -160,7 +160,7 @@ if not defined HOME set "HOME=%USERPROFILE%"
 :: This is either a env variable set by the user or the result of
 :: cmder.exe setting this variable due to a commandline argument or a "cmder here"
 if defined CMDER_START (
-    cd /d "%CMDER_START%"
+    cd /d "%CMDER_START%\"
```
daxgames: Don't assume you need to add a '\', instead add the below between 162 and 163:

if not "%CMDER_START:~-1%" == "\" SET "CMDER_START=%CMDER_START%\"

Then remove the '\' from line 163

daxgames: @Hugo-KTM Something is not right with my code figuring it out now.

daxgames: It does not like being all on one line do this instead:

if defined CMDER_START (
    if "%CMDER_START:~-1%" neq "\" (
        SET CMDER_START="%CMDER_START%\"
    )

    cd /d "%CMDER_START%"
)

Make these changes and I will approve this change.

Hugo-KTM:
Please note that I had to change the registry settings for "Drive" to make the change work:
Do not use double quotes for the %V argument for a drive (unnecessary and causes error due to trailing double quote being set in CMDER_START environment variable by Cmder.exe).

Registry
========
```reg
[... Classes\Drive\shell\Cmder\command]
@="\"C:\\Users\\<User>\\bin\\PortableApps\\cmder\\Cmder.exe\" %V"

[... Classes\Directory\shell\Cmder\command]
@="\"C:\\Users\\<User>\\bin\\PortableApps\\cmder\\Cmder.exe\" \"%V\""

[... Classes\Directory\Background\shell\Cmder\command]
@="\"C:\\Users\\<User>\\bin\\PortableApps\\cmder\\Cmder.exe\" \"%V\""
```

daxgames:
@Hugo-KTM, Thanks.

You did part of what I asked.

I like the change and it addresses an issue a user reported earlier but it is not complete in my mind. Can you please replace all instances of 'cd /d' with 'pushd' in 'vendor\init.bat' to ensure this works in all situations.

Then I will approve.

Hugo-KTM:
Sorry for that. I could not get the search I used to show me those other two instances.

It took me some time to figure out that Cmder.exe screwed up the value of ```CMDER_START``` when I gave it the root directory with double quotes and this caused all kinds of issues besides the fact that ```echo``` in cmd changes the behaviour of code around it.
I observed that ```echo``` before an ```if``` executed fine but caused subsequent ```if``` statements to fail. Remove the ```echo``` and the ```if``` works fine. Maybe using ```set``` to view variable values works better.

vendor/init.bat
```diff
@@ -149,20 +149,25 @@ call "%user-aliases%"
...
-    cd /d %USERPROFILE%
+    pushd "%USERPROFILE%"
 )
 
 :: Set home path
 if not defined HOME set "HOME=%USERPROFILE%"
 
 :: This is either a env variable set by the user or the result of
 :: cmder.exe setting this variable due to a commandline argument or a "cmder here"
+setlocal enabledelayedexpansion
```
daxgames: I added to this to do what I was asking.

daxgames:
@hugo-ktm I ended up removing my changes to your code after it was merged to cmder master.  Please pull down cmdere master and verify all still work is.

Ok. Think to truly make this work the registry changes neerd tone added to cmder.exe.  I may work on that.

Hugo-KTM:
Works for me

PR #1453, opened 2017-07-27
Doc updates

daxgames:
For profile/alias settings scripts for various shells, what loads them, and their load order.


Stanzilla:
Thanks!

PR #1456, opened 2017-07-31
allow cmder (cmd) to start in network locations

Hugo-KTM:
`PUSHD` instead of `CD /D` allows the use of UNC paths with cmd.
PUSHD temporarily maps a network drive to Z: and in this way allows cmd to with those paths.
The trailing `\` is important for the command to work in certain cases.


daxgames:
@Hugo-KTM, I like the change and it addresses an issue a user reported earlier today but it is not complete in my mind.  Can you please replace all instances of 'cd /d' with 'pushd' to ensure this works in all situations.

daxgames:
@Hugo-KTM you should just combine this with #1399.

Hugo-KTM:
integrated change into #1399

Issue #1200, opened 2016-12-14
Multiple problems when run from C:\Program Files\Cmder


brechtm:
Due to file permissions, Cmder won't start up:

    Failed to backup ConEmu.xml file to ./config folder!

After changing the file permissions, Cmder will start. But opening a mintty shell isn't possible due to spaces in the path:

    'C:\Program' is not recognized as an internal or external command,
operable program or batch file.

Jackbennett:
Oh yeh windows treats the program files area extra special, we don't recommend cmder there at all. Would you move it please and update us if the "space in path" error is still present to investigate?

brechtm:
`C:\Program Files` is the default location for applications in Windows, so it is strange that Cmder doesn't work when installed there, especially since this is not mentioned in the install instructions.

Since I cannot install it to Program Files anyway, I have now installed it to C:\Cmder, sidestepping any possible issues caused by spaces in the path.

Jackbennett:
I'm sure it used to, sorry for the run around.

Vista+ added protections to that path and UAC add more as well. We'd need a proper installer and code signing to use program files, and then have to split your customisations into appData, which for software that's supposed to be easily portable.. well, it isn't once you do that. 

You might want to use `C:\programData\cmder` for somewhere equally out the way yet for programs if you don't want cmder around your documents or something. That's untested so let us know how you get on.

dixonge:
Thanks for this documentation @brechtm !

jcaudle:
> You might want to use C:\programData\cmder for somewhere equally out the way yet for programs if you don't want cmder around your documents or something. That's untested so let us know how you get on.

I just moved over to using this location and things seem completely fine. It might be good to put this advice in the installation instructions.

daxgames:
I'm not trying to be confrontational or anything but why does so many people insist on putting this in C program files it's a portable piece of software it can go anywhere.

Cmder relies on several unix/linux like binaries where has with spaces are generally not used.

While paths with spaces in them are a common thing in Windows it was never a good idea.  Microsoft has even moved away from the practice where possible in newer versions of windows.

If it doesn't work in a specific location move it to another location.

MartiUK:
[...]

PR #1474, opened 2017-08-28
Add documentation about Program Files

jcaudle:
cmder should not be installed in `C:\Program Files` because it would need Administrator privileges to update configuration. This commit adds documentation to the installation instructions as described in #1200.


Stanzilla:
Thanks!

PR #1531, opened 2017-10-29
Fix alias.bat handling "user-aliases.cmd" with spaces

Varriount:
This fixes the alias command when user-aliases.cmd is located in a path with spaces (like when Cmder is installed into the program-files directory).
I've also made the alphabetical case of the ALIASES variable consistent.

### DIFF
--- a/bin/alias.bat
+++ b/bin/alias.bat
@@ -1,8 +1,8 @@
 @echo off
 
 
-if "%aliases%" == "" (
-  set ALIASES=%CMDER_ROOT%\config\user-aliases.cmd
+if "%ALIASES%" == "" (
+  set aliases="%CMDER_ROOT%\config\user-aliases.cmd"
 )
 
 setlocal enabledelayedexpansion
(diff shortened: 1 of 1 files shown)


bin/alias.bat
```diff
@@ -1,8 +1,8 @@
 @echo off
 
 
-if "%aliases%" == "" (
-  set ALIASES=%CMDER_ROOT%\config\user-aliases.cmd
+if "%ALIASES%" == "" (
+  set aliases="%CMDER_ROOT%\config\user-aliases.cmd"
```
MartiUK: Is this supposed to be lowercase?

Stanzilla: Yeah that looks a bit weird

Varriount: I can change it to either be upper case or lower case - Command prompt variables are case insensitive.

The original problem was that half the file was using an upper case variant of the name, while the rest was using a lower case variant.

Stanzilla: well imo it should be consistent if you change it now, change it everywhere, no?

Stanzilla: Can you change it?

PR #1611, opened 2018-01-09
jump to work directory when create new tab or split window

iflody:
when we need to create new tab or split window, we usually need to jump to work directory.

### DIFF
--- a/vendor/init.bat
+++ b/vendor/init.bat
@@ -167,6 +167,14 @@ if defined CMDER_START (
     pushd "%CMDER_START%"
 )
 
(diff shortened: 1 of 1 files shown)


MartiUK:
Hi @iflody can you rebase this to the latest master? Thanks.

daxgames:
Can't opening a new tab in the same dir be done by running `cmder /single %cd%` inside Cmder?

daxgames:
Also looking at having Cmder.exe accept unknown arguments and just passing them to Conemu as a feature. This would allow you to open a tab with a specific task in a specific folder and name it.

PR #1658, opened 2018-02-18
Detect and compare installed Git version against vendored Git

### LINKED ISSUE #1564: init.bat – Invalid method used for determining a recent `git` version
### Found old git version in "C:\path\to\git", but not using...

Sounds familiar? Upon inspecting the [`init.bat`](https://github.com/cmderdev/cmder/blob/master/vendor/init.bat#L46) file, I found out that the mechanism for checking a recent version of git is as follows:

> also check that we have a recent enough version of git (e.g. test for GIT\cmd\git.exe)

Which means that a possible installation of `git.exe` in a directory such as `GIT\bin\git.exe` is **considered NOT recent to cmder.**

This seems like a really, _rellay_ bad idea to me. I'll explain why:

Consider this for example:

cmder_git

As you can see, I have a perfectly recent `git.exe` installation within Cygwin, which happens to be located in a `/bin` directory rather than a `/cmd` directory.

However, although **`git version 2.13.2`** is obviously more recent than **`git version 2.9.0.windows.1`**, since it is not installed in a `/cmd` directory, `init.bat` thinks that it is **NOT** a recent version of git, while it obviously **is**.

While `init.bat` is a great part of cmder, I think the implementation of this part is really horrible. IMO, instead of checking for the `/cmd` path, it should compare `git.exe --version` values to detect which git is more recent.

If my proposal seems reasonable to you, I'll be glad to PR a solution, based on [this](https://stackoverflow.com/a/4025065/1454514).

Thank you :)

DRSDavidSoft:
version_compare

**Summary**
Now cmder will actually execute and compare the installed Git version against the vendored Git version, which fixes #1564 and related issues.

This is required due to unix-flavored Git installations on Windows (such as Cygwin binaries) are actually installed in a `/bin` directory rather than `/cmd`, and could even be more recent than the vendored one.

**Tests**
I've written the logic and tested the PR in the following conditions:
- No vendored binaries was deployed (Cmder minimal) → User provided used.
- No user provided binaries was available → Vendored binaries used.
- Both user and vendored binaries available:
  - The user provided binaries are more recent **OR** the same version as the vendored → User binaries used over the vendored binaries.
  - The user **actually** had an old version of git → Voilà! An older version of git found, and it's not used. Instead the vendored git is used. (I also added the older version string to the prompt, which indicates a problem that the user has to correct, either by removing or upgrading the user binaries.)

Please verify and merge the PR into main branch.
Thanks for developing such a great Terminal! :D

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,9 +147,15 @@ Note: These are loaded in this order by '$ENV:CMDER_ROOT\\vendor\\user-profile.p
 
 ### SSH Agent
 
-To start SSH agent simply call `start-ssh-agent`, which is in the `vendor/git-for-windows/cmd` folder.
+To start the vendored SSH agent simply call `start-ssh-agent`, which is in the `vendor/git-for-windows/cmd` folder.
 
 If you want to run SSH agent on startup, include the line `@call "%GIT_INSTALL_ROOT%/cmd/start-ssh-agent.cmd"` in `%CMDER_ROOT%/config/user-profile.cmd` (usually just uncomment it).
+
+### Vendored Git
+
+Cmder is by default shipped with a vendored Git installation.  On each instance of launching Cmder, an attempt is made to locate any other user provided Git binaries. Upon finding a `git.exe` binary, Cmder further compares its version against the vendored one _by executing_ it. The vendored `git.exe` binary is _only_ used when it is more recent than the user-installed one.
+
+You may use your favorite version of Git by including its path in the `%PATH%` enviroment variable.  Moreover, the **Mini** edition of Cmder (found on the [downloads page](https://github.com/cmderdev/cmder/releases)) excludes any vendored Git binaries.
 
 ### Using external Cygwin/Babun, MSys2, or Git for Windows SDK with Cmder.
 
--- a/vendor/init.bat
+++ b/vendor/init.bat
@@ -47,26 +47,54 @@ if not defined TERM set TERM=cygwin
 :: * if the users points as to a specific git, use that
 :: * test if a git is in path and if yes, use that
 :: * last, use our vendored git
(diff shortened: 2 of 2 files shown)


DRSDavidSoft:
@Stanzilla I'm confused. This path is approved but not merged?

Stanzilla:
Yeah I'm not alone here, waiting for others to chime in. @cmderdev/trusted-contributors

daxgames:
I like it as long as the behaviour fully is documented in readme.md  I am no it where I can test it though.

DRSDavidSoft:
@daxgames Added a full section regarding the changes in `README.md`. Could you confirm?

Jackbennett:
Looks great thanks for the contribution.

daxgames:
I am testing this after merge to master and it fails to find a valid git version on vendored git.  Problem appears to be here:

    :: get the git version in the provided directory
    for /F "delims=" %%F in ('%git_executable% --version 2^>nul') do @(
        set "GIT_VERSION_%~1=%%F"
        call :verbose-output GIT_VERSION_%~1=%%F
    )

Vendored git returns a version string without '=' so no version is ever set.

Trying to fix it in the PR I have been working on.

daxgames:
I do not use cygwin.  what is returned by `git --version` in cygwin?

Vendored git returns `git version 2.15.0.windows.1`

daxgames:
FIxed in #1665

DRSDavidSoft:
@daxgames Cygwin (or any unix falvored build of Git) returns something in the lines of: `git version 2.15.0.1`.

I tested it out before pushing it to master, and it was able to return and parse both the vendored and user binaries.

What was the verboose output when you posted this? Is the issue fixed now?

PR #1665, opened 2018-02-23
New cmder.exe args and shared install capability

daxgames:
## [1.3.6-pre1](https://github.com/cmderdev/cmder) (2018-03-01)

**Fixed bugs:**

* Fixed Git version check recently added to master.

**Updates:**

* Modified Cmder tasks in default Conemu.xml to allow easily adding command line args for init.bat by adding some quotes.  This resulted in a ton of misc changes to this file.  See Adds below.
* Reworked `cmder.exe` command line argument handling to make it more flexible and easily added to.
* Reworked README.md tables to make them more readable in editors

**Implemented enhancements:**

* Append `%GIT_INSTALL_ROOT%\mingw32|64` folder to the path if it exists. Thanks: @gucong3000 
* Added `cmder.exe` command line args documenttion to `README.md`
* Added `:enhance_path` method to vendor\init.bat that modifies the path only if required.
   * To prepend: `call :enhance_path "%cmder_root%"`
   * to append: `call :enhance_path "%cmder_root%" append`
* Added `:enhance_path_recursive` method to vendor\init.bat that adds a path and all its sub directories to the path if required.  
   * Max recurse depth default is '1' configurable using `init.bat /max_depth [1-5]`. 6+ results in error. 
   * To prepend and go 3 levels deep: `call :enhance_path "%cmder_root%" 3`
   * To append and go 2 levels deep: `call :enhance_path "%cmder_root%" 2 append`
* Added ability to init.bat to accept command line args and documented them in README.md.  Allows users to change the behaviour of init.bat without editing the file.

| Argument                      | Description                                                                                      | Default                               |
| ----------------------------- | ----------------------------------------------------------------------------------------------   | ------------------------------------- |
| /c [user cmder root]          | Enables user bin and config folders for 'Cmder as admin' sessions due to non-shared environment. | not set                               |
| /d                            | Enables debug output.                                                                            | not set                               |
| /git_install_root [file path] | User specified Git installation root path.                                                       | '%CMDER_ROOT%\vendor\Git-for-Windows' |
| /home [home folder]           | User specified folder path to set `%HOME%` environment variable.                                 | '%userprofile%'                       |
| /max_depth [1-5]              | Define max recurse depth when adding to the path for `%cmder_root%\bin` and `%cmder_user_bin%`   | 1                                     |
[...]

PR #1683, opened 2018-03-01
Lib it

daxgames:
# Merge #1665 first.

* Waiting on @DRSDavidSoft to contribute the lib_git.cmd per jis request.

## [1.3.6-pre2](https://github.com/cmderdev/cmder) (2018-03-01)

**Updates:**

* Removed all sub routines from `init.bat` and made them into importable libraries that can be used in any `*.bat|cmd` file.

  * Libraries are in `%cmder_root%\vendor\lib`
  * Import libraries into any `*.bat|cmd` file using `call "%cmder_root%\vendor\lib\[library file name]"`
  * Call library methods by typing `"%lib_path% enhance_path "c:\bin"`
  * Get help on library method usage by typing `"%cmder_root%\vendor\lib\[full library file name]" /h`


DRSDavidSoft:
@daxgames Thanks for the wait, will let you know in daxgames/cmder/pull/4 when the git subroutine PR's ready to be merged.

daxgames:
@DRSDavidSoft Please remove the @ signs before any ( in your for blocks.

DRSDavidSoft:
@daxgames Why closed?

DRSDavidSoft:
@daxgames Also curious to know, why you removed readline settings from config dir?

daxgames:
@DRSDavidSoft closed becasue I need to rebase it after `init_args` is merged the commit history is a mess since it was derived from  `init_args` and I will reopen it then.

Sorry thought the `settings` in the config dir was a mistake.  But since you asked :-)

The `%cmder_root%\config` dir should have nothing in it by default other than the default `conemu,xml` and a readme.md because everything in this dir is supposed to be user owned and controlled.  The `conemu.xml` should not even need to be there just that no one has moved it to `%cmder_root%\vendor` where it should be.

We want to lay down a `%cmder_root%\config\settings` file exactly one time for the user and from then on it is their config.  We do not want to overwrite it on an update of cmder by copying a newly released zip contents into an old directory.

The default settings file is generated the first time `clink.exe` loads.  If the default settings file is sub optimal the I suggest handling it the same way we do other files like `user-aliases.cmd`.  If it does not exist drop it there from a sample file in `%cmder_root%\vendor` before `clink.exe` runs.

daxgames:
@DRSDavidSoft Also need to to test this again.  I made some changes to git version handling that I found issues for.  Want to make sure it still works for your use case.

DRSDavidSoft:
@daxgames Thanks for the great explanation, I did not realize that the `/config` directory is for user's config. Which also means that the `ConEmu.xml` should be moved to the `/vendor` directory next to init.bat (though I vote to separate vendor binaries and vendor configs from each other.) That should also mean that Cmder could also work with a non-vendored installation of ConEmu.

Since you mentioned it, is #1665 ready to merge, or as you said it requires further testing?

daxgames:
I be live #1665 is ready to merge but would rather someone test the nightly build other than me.cmder.exe args were severely broken before today in this branch.

To clarify the conemu.xml in config is just the default conemu config.  It is never used after the first cmder launch.  The actual file conemu uses is in the conemu folder and that file is copied  to the config folder each time cmder starts to preserve user changes post upgrade.

I would not count on cmder working 100%  with a non-vendored conemu, it might, but things might be missing.

PR #1690, opened 2018-03-06
Add MinGW commands to path

gucong3000:


DRSDavidSoft:
@daxgames This seems reasonable, but against the "Lib it" branch

daxgames:
Not sure about this one.  Need to test it and and decide.  Probably will get added just not exactly like this.  Holding off for now to let some of the dust settle.

DRSDavidSoft:
@daxgames @gucong3000 Also MinGW should **only** be added to path, if _vendored_ Git is decided to be used by `init.bat`.

daxgames:
@DRSDavidSoft Vendored Git or user installed 'Git for Windows'.  Notice the code only adds it if the folder exists.

daxgames:
@gucong3000 - I am adding this to my PR #1665 as it already has enhancements to how we update the path by adding this to my PR I can do it by adding 4 lines of code:

```
    if exist "!GIT_INSTALL_ROOT!\mingw32" (
        call :enhance_path "!GIT_INSTALL_ROOT!\mingw32" append
    ) else if exist "!GIT_INSTALL_ROOT!\mingw64"
        call :enhance_path "!GIT_INSTALL_ROOT!\mingw64" append
    )
```

PR #1692, opened 2018-03-06
simplified λ prompt

DRSDavidSoft:
The lambda (λ) character for the prompt is defined only once instead of twice, making it easy for customizations to change shell handle character.


vendor/clink.lua
```diff
@@ -40,12 +40,10 @@ local function set_prompt_filter()
     -- orig: $E[1;32;40m$P$S{git}{hg}$S$_$E[1;30;40m{lamb}$S$E[0m
     -- color codes: "\x1b[1;37;40m"
     local cmder_prompt = "\x1b[1;32;40m{cwd} {git}{hg}{svn} \n\x1b[1;39;40m{lamb} \x1b[0m"
-    local lambda
+    local lambda = "$"
```
daxgames: Doesn't this replace the "λ" with "$"?  Or am I missing something?

DRSDavidSoft: @daxgames Oh sorry, an invalid commit made its way to this branch. I'll make a new PR

DRSDavidSoft: @daxgames fixed in #1693. Sorry for sloppy git work

Issue #1688, opened 2018-03-03
Custom cmder prompt


victormongi:
How to make look like this?
![image](https://user-images.githubusercontent.com/5617746/36933713-f1fb00b4-1f18-11e8-82ae-15a760e50759.png)
Thank you very much...

gucong3000:
https://github.com/cmderdev/cmder/blob/master/vendor/clink.lua#L50
```lua
    local cmder_prompt = "\x1b[1;32;40m{cwd} {git}{hg}{svn} \n\x1b[1;39;40m{lamb} \x1b[0m"
    local lambda
    cmder_prompt = string.gsub(cmder_prompt, "{cwd}", cwd)
    if env == nil then
        lambda = "λ"
    else
        lambda = "("..env..") λ"
    end
    clink.prompt.value = string.gsub(cmder_prompt, "{lamb}", lambda)
```

victormongi:
This is my example code
```
    local cmder_prompt = "\x1b[1;32;1;40m{lamb} \x1b[36m"..dir.." \x1b[32;1;1m{git}{hg}{svn} \x1b[0m"
    local lambda

    -- char set • →
    cmder_prompt = string.gsub(cmder_prompt, "{cwd}", cwd)
    if env == nil then
        lambda = "→"
    else
        lambda = "("..env..") →"
    end
    clink.prompt.value = string.gsub(cmder_prompt, "{lamb}", lambda
```
![image](https://user-images.githubusercontent.com/5617746/37023273-52eab862-2160-11e8-81cc-b77a17fdf8ad.png)

gucong3000:
Mabe here: https://github.com/cmderdev/cmder/blob/8237b1010cbedefb57850144d55ea36c5fbee0c3/vendor/clink.lua#L266-L272

victormongi:
Hi @gucong3000 Ive been trying to achieve this.
But how to make the prompt like this `→ folderName git(branchname)` instead of `→ folderName git(branchname -> origin)`. Where to remove the `-> origin`?
Thank you...

daxgames:
@victormongi The easiest way while not really correct would be to delete the file '%cmder-root%\vendor\clink-completions\git_prompt.lua'.

I am working on a more corrwct way to do this.

victormongi:
Thank you @daxgames it just work...
![image](https://user-images.githubusercontent.com/5617746/37071678-a2500a9a-21f8-11e8-8b32-012eb53dde01.png)
[...]

Issue #1261, opened 2017-02-18
Customising Lambda


justskymarshall:
Is the only non-customisable component of cmder, the lambda itself? I'd like it to be a different character.

daxgames:
You could customize it but you would need to redo the customization after each upgrade.  It is not designs to be user configurable.

gustavoalfaro:
I have my own customized theme for cmder you can find it here:
https://github.com/gustavoalfaro/power-monokai

Jackbennett:
We'll work out a solution to #1280 and you'll be able to change everything.

Issue #104, opened 2013-12-30
Changed "λ"  to "?"


AndreyUtka:
When I go to powershell, symbol "λ" changed to "?".
![2013-12-30_1837](https://f.cloud.github.com/assets/1304912/1823052/579f727e-7171-11e3-920e-0c6e4f36db1a.png)

f3yisayo:
@bravelee I also wanted to have this instead of the Lambda, I did some digging and found a Lua script in `CMDER\config\` 

1. Open the `prompt.lua` script in the path above in an editor.
2. On the second line, just replace `λ` with `$` and save, have fun.

haoxi911:
@f3yisayo I didn't see the prompt.lua script in the latest version, is there another way to replace `λ` with `$`? I like the tool but it looks weird to type something after `λ`.

Ov4ts:
@haoxi911 I found a new way, go to ~cmder\vendor, open clink.lua, line 44, change to (lambda = "$") and save. Done.

haoxi911:
@Ov4ts It works great. Thank You!

taichushouwang:
@Ov4ts What you say make cmd work, but for the powershell, it's still λ

mataps:
to change for λ replace the ff files:
`cmder\vendor\clink.lua`  
`cmder\vendor\profile.ps1`  
`cmder\vendor\git-for-windows\etc\profile.d\git-prompt.sh`

smltrs0:
UPDATED

CMDER\config\_powerline_config.lua

--
plc_prompt_lambSymbol = "NEW SIMBOL"

taowaaCool:
update:
edit .../cmder/config/cmder_prompt_config.lua
...
prompt_lambSymbol = "NEW Symbol"
...

PR #1693, opened 2018-03-06
simplified λ prompt in lua

DRSDavidSoft:
The lambda (λ) character for the prompt is defined only once instead of twice, making it easy for customizations to change shell handle character.


DRSDavidSoft:
@daxgames You might want to squash and merge this, because of the sloppy commits.

DRSDavidSoft:
Related issues: #1688 #1261 #104 
Also on another note, I know λ is an iconic part of Cmder, but an option for the user to customize it would be nice.

DRSDavidSoft:
@Stanzilla is this able to merge? I want to work on  another branch to customize the prompt character and color.

Stanzilla:
Yup, looks good, thanks for that! Wonder why I never changed it...

PR #1725, opened 2018-03-23
open cmder in the current working dir as every other windows app works

daxgames:
Opens cmder in the current working directory unless `[path]` or `/start [path]` arg is used.


daxgames:
@MartiUK will do,was just testing the waters did not know how this would be received.

PR #1754, opened 2018-04-26
Fix git install root

daxgames:
This is untested so don't merge yet.  I'll test tonight.


Stanzilla:
Status?

daxgames:
Ready to go! Tested and looks like it works to me.

PR #1758, opened 2018-04-30
lib to check init.bat's custom args 

xiazeyu:
This commit deals with the issue when I tried to use `init.bat` for both `cmder.exe` and `Code.exe` (VS Code).
in the user-profile, I set an Autorun function:
```batch
set Option1=%1
if not "%Option1%" == "noAutoRun" (
  echo Auto run
  call devDocs
  call SublimeText
  call vsCode
)
```
In `cmder.exe`, these applications will be run once I start `cmder.exe`,
and will **not run** when I use it in `VSCode`
```json
"terminal.integrated.shell.windows": "cmd.exe",
  "terminal.integrated.shellArgs.windows": [
    "/k",
    "%cmder_root%/vendor/init.bat",
    "noAutoRun"
  ]
```

to achieve this, I passed the arguments received by `init.bat` to `user-profile.cmd` and it seems OK.


xiazeyu:
@daxgames Sorry that I've missed `init.bat`'s built-in command parser
Now I've fixed it by using %CMDER_FLAGS% and added notes in `user-profile.cmd`

By the way, it seems that v1.3.5 isn't match current branch at all. Is that OK?

daxgames:
@xiazeyu - Yes 1.3.5 is the released version of cmder.  the master branch has unreleased changes.  Also I made a couple more small changes just to make things a little clearer.

@cmderdev/trusted-contributors do you approve of this change?

vendor/init.bat
```diff
@@ -317,6 +320,19 @@ echo.
 echo :: you can add your plugins to the cmder path like so
 echo :: set "PATH=%%CMDER_ROOT%%\vendor\whatever;%%PATH%%"
 echo.
+echo :: Pass custom arguments to init.bat, and you can parse them like so:
+echo :: If found...
+echo :: echo %* | find /i "/noautorun">nul
+echo :: if "%ERRORLEVEL%" == "0" (
+echo ::   call vsCode
```
DRSDavidSoft: is `vsCode` a batch file, or a subroutine? Why is it called here?

daxgames: It's not it's a comment

xiazeyu:
@daxgames @DRSDavidSoft I've optimize the comments about this feature, have a look plz.

Stanzilla:
@daxgames ready to merge or nah?

DRSDavidSoft:
### P.S. 1

@daxgames I can write a `lib_arguments.bat` to parse the arguments, If it's okay by you. It should parse these types of switches IMO:
* `/switch`
* `--switch`
* `-s` and/or `/s`

### P.S. 2

@xiazeyu On another note, we have a Wiki page explaining on how to integrate Cmder with VS Code, here: https://github.com/cmderdev/cmder/wiki/Seamless-VS-Code-Integration

If you have something useful in your `vsCode.bat`, we'd appreciate posting it there.

Here's also the wiki page for Sublime 3 integration: https://github.com/cmderdev/cmder/wiki/Seamless-Sublime-Text-3-Integration

daxgames:
@drsdavidsoft I don't understand the argument handling question and I don't understand the plugins comment because there is no argument handling because any unknown argument to is just passed to user-profile.cmd and there are no plugins so can you explain?

DRSDavidSoft:
@daxgames Sure. I propose adding a _simple_ wrapper to be used as argument handler, instead of `find /i`'ing and comparing the produced error level each time. Something like this:

```
REM place `user-plugins.bat` functions here

:exampleFunction
    echo. Something is executed!
    exit /b

:anotherFunction
    echo. Something else is being executed!
    exit /b

REM this section should be called from either `init.bat` or `user-profile.bat`

call :conditionalCall switchname1 exampleFunction
call :conditionalCall switchname2 anotherFunction

REM this section goes into `lib_args.bat` or `lib_base.bat`, etc

:conditionalCall 
    set arg_name=%1
    set subroutine_name=%2

    call :matchArgument %arg_name% %CMDER_FLAGS%
    if "%ERRORLEVEL%" == "0" call :%subroutine_name%

    exit /b
```

Of course, this is just an idea. If this doesn't make any sense, let's ommit it and stick to parsing arguments directly in `user-profile.bat`

My other main point (regarding documentation) is that this functionality (and how to use it with `user-profile.bat`) should be addressed in `README.md`.

xiazeyu:
@DRSDavidSoft 

I think document this function in `README.md` is a good idea.

In my `vsCode.cmd`, there's nothing but a `call %cmder_root%/bin/vsCode/code.exe`. I use cmder as an entry point of my working environment, so I'd better to have some commands to shell these editors with UI, not intergrade cmder into them, and that's what my `vsCode.cmd` for. Type `vsCode`, and I can shell vs cofe, and not interrupt my current console job.

I think your wapper idea is excellent, that will make `user-profile.cmd` much better readable and clear. It's a good alternative to simple switch jobs(it can't replace those more complicated jobs which may process more than 1 argument eg. `/setEnv development`).

I also think it will be too much complex if `uder-profile.cmd` has too much, and add more file may make it bloated.

I will try if I can put this function into something like `have.cmd`, and it will just work similarly to the useage of `alias.cmd`. How ever **automatically pass %CMDER_FLAGS% to have.cmd** and **ability to call sub functions(like your `:exampleFunction` and `:anotherFunction` in `user-profile.cmd`) needs some research.

DRSDavidSoft:
@xiazeyu Glad to hear that, so I'll wait to see what you'll commit for the mentioned subroutines.
Also, could you make the README commit in this same PR?

If you need help, you can count on me. I think this functionality will be great in Cmder! :)

xiazeyu:
@DRSDavidSoft I've finished my work, and have a look.
I tried my best to test but to err is human, with your help it will be more stable.
I will update the Wiki page when the PR is merged.

daxgames:
@xiazeyu  Cmder has a concept of importable libraries in `vendor\lib` this is what @DRSDavidSoft was referring to when he suggested a 'lib_arguments.bat to parse the arguments'.  I like what you have done mostly.  I am hoping you can refactor your code using the files in `vendor\lib` as examples.

We put the libraries in `vendor\lib` because they should not be edited by users of `cmder`.  Your `have` method if it is to be a part of `cmder` should not be edited by users so it should not be in `%cmder_root%\bin`.  Eventually I would even like to move the `alias.bat` out of `%cmder_root%\bin` into a library in `vendor\lib` for this same reason.

I am not sure I like `have` as the `method` name, I am thinking `flag_exists` as the method that checks if a flag is set and `flag_exists not` could check for the absence of the flag. 

Thoughts?

xiazeyu:
@daxgames See the latest commit.
I agree with you to put these kind of scripts into `vendor\lib` to prevent users to edit them.
However, I think always using `%flag_exists% xxx xxx` is not that concise, maybe alias it is or put it into `%PATH%` a better idea, since users **don't use %lib_git% quite often**, but %flag_exists% is **directly used by users**.

I'm poor at naming functions, so I really think flag_exists is a better idea.

DRSDavidSoft:
@daxgames, thoughts?

@xiazeyu I'm reviewing the PR, if everything's alright I'll ask @Stanzilla to merge it.

Thanks for the contribution, btw!

xiazeyu:
@DRSDavidSoft @daxgames @Stanzilla 
By the way, is it possible to let me know when will the next stable version be released?

daxgames:
We just released 1.3.6.  It has a couple of issues that have already been fixed.  I think we might need to do another one sooner rather than later.

daxgames:
Still have not had time to look at this PR.  I need to before we merge it, something doesn't feel right about it still but I can't say for sure until I have time to really sit down,and look at it.

xiazeyu:
Since vsCode official supported the "Portable Mode"
https://code.visualstudio.com/docs/editor/portable

These documents may need to be modied.

xiazeyu:
Instructions about Integration with vsCode have been migrated to https://github.com/cmderdev/cmder/wiki/Seamless-VS-Code-Integration

Stanzilla:
@daxgames had time to take a look yet?

xiazeyu:
@daxgames I haven't found anything unexpected till now. I think there's a low possibility that this new feature would go wrong after configured correctly.

daxgames:
@stanzilla no I haven't, work has been very busy.

@xiazeyu explain 'there's a low possibility that this new feature would go wrong after configured correctly'

xiazeyu:
@daxgames Reasons are as follows:
1. This new feature has **nothing to do with the kernal of cmder**. It **won't be executed after user-profiled.cmd** unless you call it. And it is designed to be **executed optionally only at startup**.
2. The only API it has changed is **arguments passed to `init.bat`**. In previous version, init.bat was called by ConEmu **with no arguments**. Even if you passed arguments to `init.bat` accidentally, without the `%%flag_exists%%` command, this function won't be executed.
3. So the only problem it may have is **flag_exists.cmd itself** and **the way user call it (or its configuration)**.

- flag_exists.cmd: I've been tested it for 2 months, and it showed no unexpection. Since the command of executing `init.bat` is a kind of fixed command, not dynamic, I don't think it will have any strange cases.
- User configuration: because the useage of this feature is simple, and the only two unexpected behaviour are *want to execute but won't work* and *don't want to execute but executed*. In the worst situation, all these two problems can be solved simplely by **not using this function**. Moreover, it is easy to test configuration by run the program several times and change the configuration. Our document provided **all kinds of examples**, so it makes it easier to learn this new feature.

Thats all my explanations.

xiazeyu:
@daxgames @DRSDavidSoft I've found some problems.
https://github.com/cmderdev/cmder/pull/1758/commits/f6c2d9c31e86666cae35ff17911c5495452b587b#diff-cabfc4f102e08adb66d90c726994933fR10
I'm not sure to where to declare `endlocal` so I put it here, but I found that once I use `setlocal`, the `init.bat` won't behave properly.

vendor/init.bat
```diff
@@ -7,10 +7,11 @@
 :: !!! Use "%CMDER_ROOT%\config\user-profile.cmd" to add your own startup commands
 
 :: Use /v command line arg or set to > 0 for verbose output to aid in debugging.
+endlocal
```
xiazeyu: Here's the strange point. If I use
```batch
endlocal
set "CMDER_USER_FLAG= "
setlocal
```
the whole `init.bat` won't behave properly.

vendor/init.bat
```diff
@@ -7,10 +7,11 @@
...
 set verbose-output=0
 set debug-output=0
 set max_depth=1
-set "CMDER_FLAGS="
```
xiazeyu: Migrate `CMDER_FLAGS` into **global varible** `CMDER_USER_FLAGS`

vendor/init.bat
```diff
@@ -312,8 +313,7 @@ echo :: use in front of the command to prevent printing the command
 echo.
 echo :: the next two lines is for "%%flag_exists%%" shortcut, a custom arguments handler
 echo :: don't remove it if you need it
-echo set "CMDER_USER_FLAGS=%%*"
-echo call "%%cmder_root%%\vendor\lib\flag_exists"
+echo call "%%cmder_root%%\vendor\lib\flag_exists" "/setPath"
```
xiazeyu: I'm not sure if this `/setPath` is good or the old one(add global varible `%flag_exists%` automatically) is better.

vendor/init.bat
xiazeyu: Also I'm not sure if %GIT_INSTALL_ROOT% goes wrong have something to do with this.
I run `echo %GIT_INSTALL_ROOT%` and the output is `!GIT_INSTALL_ROOT!`. And I'm quite confused.

Stanzilla:
@daxgames is your review still valid?

daxgames:
I don't even remember reviewing it.  Last I saw the author had some concerns.  I have not really looked at it.

xiazeyu:
These concerns are just about some style issue, and since it can work, I don't think it matters a lot.

README.md
```diff
@@ -246,7 +247,42 @@ Uncomment and edit the below line in the script to use Cmder config even when la
 # CMDER_ROOT=${USERPROFILE}/cmder  # This is not required if launched from Cmder.
 ```
 
-### Integrating Cmder with [Hyper](https://github.com/zeit/hyper), [Microsoft VS Code](https://code.visualstudio.com/), and your favorite IDEs 
+### Handling with custom arguments when using init.bat
```
DRSDavidSoft: Since the actual flag comparison should be done in `/config/user-profile.cmd` and not `init.bat`, I suggest not mentioning `init.bat` as the arguments handler; instead advertise `user-profile.cmd` as where the arguments are handled.

README.md
```diff
@@ -138,16 +138,17 @@ You may find some Monokai color schemes for mintty to match Cmder [here](https:/
...
 | `/max_depth [1-5]`              | Define max recurse depth when adding to the path for `%cmder_root%\bin` and `%cmder_user_bin%`   | 1                                     |
-| `/svn_ssh [path to ssh.exe]`    | Define `%SVN_SSH%` so we can use git svn with ssh svn repositories.                                | `%GIT_INSTALL_ROOT%\bin\ssh.exe`      |
+| `/svn_ssh [path to ssh.exe]`    | Define `%SVN_SSH%` so we can use git svn with ssh svn repositories.                              | `%GIT_INSTALL_ROOT%\bin\ssh.exe`      |
 | `/user_aliases [file path]`     | File path pointing to user aliases.                                                              | `%CMDER_ROOT%\config\user-liases.cmd` |
 | `/v`                            | Enables verbose output.                                                                          | not set                               |
+| (custom arguments)              | User defined arguments processed by `flag_exists`. Type `%flag_exists% /?` for more useage.      | not set                               |
```
DRSDavidSoft: I'm not sure how the launcher (`Cmder.exe`) passes arguments to `init.bat`. Are you also launching Cmder with custom flags like this: `Cmder.exe /noAutorun` or just `vendor\init.bat /noAutorun`?

README.md
```diff
@@ -246,7 +247,42 @@ Uncomment and edit the below line in the script to use Cmder config even when la
...
+
+You can pass custom arguments to `init.bat` and use `%flag_exists%` to detect it.
+
+It is useful when you have multiple modes to execute cmder.
+
+If you use
+
+```batch
+
+init.bat /startNotepad
```
DRSDavidSoft: Is using `init.bat /startNotepad` the only method, or does `Cmder.exe /startNotepad` also work?

vendor/init.bat
```diff
@@ -312,18 +316,27 @@ if not exist "%initialConfig%" (
     echo Creating user startup file: "%initialConfig%"
     (
 echo :: use this file to run your own startup commands
-echo :: use  in front of the command to prevent printing the command
+echo :: use in front of the command to prevent printing the command
+echo.
+echo :: the next two lines is for "%%flag_exists%%" shortcut, a custom arguments handler
+echo :: don't remove it if you need it
```
DRSDavidSoft: Since `init.bat` should be left unchanged by the users, I doubt they'll edit this file. In any case, we should discourage users from directly editing this file.

vendor/lib/flag_exists.cmd
```diff
@@ -0,0 +1,131 @@
...
+if not defined CMDER_USER_FLAGS (
+  :: in case nothing was passed to %CMDER_USER_FLAGS%
+  set "CMDER_USER_FLAGS= "
+)
+
+set "feNot=false"
+goto :parseArgument
+
+:doShift
+  shift
```
DRSDavidSoft: I'm curious, why do you call `shift` in a dedicated subroutine `doShift`?

DRSDavidSoft:
@xiazeyu Please have a look at my review; once you reply them I can make a PR to your branch to address them.

Some additional points:
- If the flags are not passed from the launcher (`Cmder.exe`), I'll write the required code to make it happen. I think it will be useful to have both launcher and `init.bat` able to set flags.
- A better suggestion for the title: **"Passing custom flags to Cmder and handling them in `user-profile.cmd`"**
- Also, I think this feature deserves its own page in the Wiki, instead of a section in the VS Code. (I know you mainly use it in combination with VS Code, but it can also be used in combination with Sublime.)
- I think the more of the docs and help/usage be placed in the Wiki is the better.

README.md
xiazeyu: The original intention of this function is to handling different entries with different behaviors.
`/config/user-profile.cmd` is one of the entry. The other entries (eg. vsCode) need to pass arguments to init.bat to get the specific environment with specific behavior.

Graph: 
`/config/user-profile.cmd` with `argA` (=> init.bat (hidden for user I think)) => cmder environment (with Behavior A)
`vsCode internal terminal` with `argB`=> init.bat => **shell** environment (with Behavior B)
`other` with `argC` => init.bat => **shell** environment (with Behavior C)

README.md
xiazeyu: I use `vendor\init.bat /noAutorun`.
it is actually Conemu called init.bat **with arguments which have nothing to do with custom flags**.
This section in README.md is **Command Line Arguments for `init.bat`**. So it **has nothing to do** with the launcher, that's why I put it here instead of `Cmder.exe Command Line Arguments`.

README.md
xiazeyu: `Cmder.exe /startNotepad` **DO NOT WORK.**
See https://github.com/cmderdev/cmder/pull/1758#discussion_r217874087

vendor/init.bat
xiazeyu: Yeah, but I think it has nothing to do with this PR.
In my point of view, this PR just **isn't** a BREAK CHANGE.
They can find this new feature in CHANGELOG or README.md.

We just recommend them not to edit `init.bat`, if they modified `init.bat`, it's their responsibility to check carefully while upgrading.

About *we should discourage users from directly editing this file.*. I don't think it has something to do with this PR.

vendor/lib/flag_exists.cmd
xiazeyu: https://github.com/cmderdev/cmder/blob/0d2e16c6536495c9ebc2466b3a15582a9aee45e3/vendor/bin/alias.cmd#L17
https://github.com/cmderdev/cmder/pull/1758/files/39f47870a339c99e7526920ec3e06b875db8890b#diff-6584be369e2738104e1069789ea6bfe4R29

For the detection of `NOT`, copied from your `alias.cmd`.

xiazeyu:
I passed to `init.bat` because I need to use it **with no launcher** in vsCode.
things like **cmder.exe /cpp** or **cmder.exe /nodejs** or  **cmder.exe /python** may be useful to those who need to initialize multi-environments but are to slow to start up.
I don't know if Sublime also have internal terminal or othre things. Things like wiki(or documents) you can write as you want. Since I recently don't have a lot of time to dealing with such documents.(also not good at it).

DRSDavidSoft:
@xiazeyu Thanks for the quick replies.

I'll commit my suggested edits and then make a PR into your branch, so you can accept it there before this PR is merged.

I see you have updated with the master, that's good for the edits.

Might I suggest making this new flag feature work when called from **both** `init.bat` and `Cmder.exe` to make it unified? Personally I only use `init.bat`, but that's for power users like you and me. The regular users will stick to `Cmder.exe` and I think adding your feature _also_ to the main launcher will greatly benefit them.

Also, @daxgames, your thoughts please.

xiazeyu:
I'm totally agree with you.
Look forward to your commits.

README.md
daxgames: @xiazeyu I am utterly confused by your example.

README.md
daxgames: The way I understand this functionality is it is to pass args to init.bat via different task entries in the conemu gui it has nothing to do with the cmder.exe launcher.

daxgames:
@DRSDavidSoft and @xiazeyu I am not sure we need to add this functionality to `cmder.exe`, I am not sure I understand why its needed.  The stated use case is to run `init.bat /noautorun` in a shell from another tool not running `cmder.exe /noautorun` standalone, both provide completely different results.  One starts a Cmder/Clink enhanced shell in the current shell session the other launches Conemu and starts a completely new shell session in the Conemu window.

I have said before I am concerned with adding so much functionality to the launcher that it becomes difficult to maintain.

I am also concerned about breaking existing functionality.

Currently `cmder.exe` accepts a start directory in the following ways:

```
cmder /start c:\start_in_this_dir
```

or

```
cmder c:\start_in_this_dir
```

We can't just throw a random arg at it because it errors on unknown/invalid command line args.

If this functionality is added it would need to be in a way that used a prefix like `/uflags "/noautorun /dosomething /dosomethingelse"`.  The value of `/uflags` could then be set to `%CMDER_USER_FLAGS%`.

To me the meat of this PR is the library that makes it easy for easy determination of whether a flag was passed or not.  Currently the lib does not appear to adhere to the format of our other libraries in `%cmder_root%\vendor\lib` so that is an issue I have.

I need to look closer though and do a full test and review of the PR, if I am misunderstanding something please clarify.

README.md
xiazeyu: @daxgames Your understanding in https://github.com/cmderdev/cmder/pull/1758#discussion_r217885586 is correct. But it's just the current design.

@DRSDavidSoft suggests an further idea that args can be passed BOTH via init.bat and via cmder.exe(the launcher, he said he could have a try.), instead of only parse args passed to init.bat. 
Which means passing arguments to the launcher is also a way to start from different entries. (instead of only by init.bat)

xiazeyu:
@daxgames I think your understanding is exactly the same with me.

About pass args to init.bat or launcher: 

> * If the flags are not passed from the launcher (`Cmder.exe`), I'll write the required code to make it happen. I think it will be useful to have both launcher and `init.bat` able to set flags.

https://github.com/cmderdev/cmder/pull/1758#issuecomment-421493343
This has nothing to do with the original design. It **depends on you** if you will *add the function of  passing args by launcher*. Since this is beyond by ability, if you think it too risky to *add such function to launcher*, just ignore the idea or mark it as a new feature and solve it in a later PR.

About format: 

I think some of the difference exists because scripts in `%cmder_root%\vendor\lib` are mostly designed for `init.bat`, and not for the user. So it won't handling with args like "/h" and wrong syntax.

If you stick to it, you need to unite the format by yourself(sorry about that), since I can't get the rule of format(stylistic issues, comments style, best process flow as a lib).(It's think this solving this issue is not like using ESLint, which has a fixed rules to follow).

daxgames:
I have adapted it to a cmder library and can PR back to @xiazeyu.

DRSDavidSoft:
@daxgames 

### Regarding the launcher arguments

To be clear, I didn't plan to throwing random arguments from the launcher to `init.bat`! I was certainly thinking of the prefixing idea, but TBH I was planning on calling it `/initflags`, but I like your `/uflags` idea better.

On the note that is it necessary or not, I would imagine as the launcher, it should be possible to pass any `init.bat` arguments using the launcher. The only argument is on _how_ to do it. If you agree on prefixing it, I'd like to ask permission to add it to the launcher.

### Regarding the format

I had planned on refactoring @xiazeyu's batch file to match the other libs (like you initially did), but considering the format and the nature of the file, I also agree with separating it from the other libs.

May I ask why `cexec.cmd`? What does that stand for?

### Regarding maintainability 

I made a note at the end of https://github.com/cmderdev/cmder/pull/1873#issuecomment-421656666

daxgames:
I see found what cexec was. I just think it is much more accurate than flag_exists.  I would expect flag_exists to simply return a binary yes or no but it does much more than that.

The only way I can see the launcher passing flags to init.bat is trough an environment variable like %CMDER_USER_FLAGS%.  Init.bat should take whatever is set add to it.

daxgames:
The only issue with the above approach is it is a manual change to the cmder task config for existing users that already have a `user-conemu.xml` file.  

Just spit balling here but we could instead add some code to `init.bat` to append `%*` to `%CMDER_ARGS%` then have the command line parser parse `%CMDER_ARGS%` instead of `%*` then it would not have to be added the the cmder task in conemu.

xiazeyu:
> Renamed `%cmder_root%/vendor/lib/flag_exists.cmd` to `%cmder_root%/vendor/bin/cexec.cmd`
> * This makes the file easily callable by its name since `%cmder_root%/vendor/bin/` is added to the path

Does it mean that you can use `cexec /startnotepad start notepad.exe` instead of `%cexec% /startnotepad start notepad.exe`? Since this `vendor/bin` didn't occur in the current release, I'm not sure about this.
Anyway, Is it a conflict or you wants to preserve both calling method for compatibility?

PR #1761, opened 2018-05-01
Plugin support updated

daxgames:
### Updated from closed PR #849 see here for history.

Added support to extend cmder using plugins in the form of json files.


.gitignore
```diff
@@ -14,6 +14,7 @@ build/
 Version v*
 *.bak
 config/user-*
+bin/utils.ps1
```
DRSDavidSoft: why is `utils.ps1` ignored? This means we couldn't push the file from local repo

DRSDavidSoft:
@daxgames +1, also @Stanzilla may I suggest adding this to the `v1.4` release project?

.gitignore
daxgames: because it is put in place by build.ps1 at build time.

DRSDavidSoft: @daxgames My bad, I had it confused with `scripts/utils.ps1` instead of `bin/utils.ps1`.

daxgames:
@DRSDavidSoft I like your enthusiasm but this PR has a history.  I like the functionality and I think others will also but the @cmderdev/trusted-contributors had some reservations.

DRSDavidSoft:
@daxgames I just hope those reservations are resolved by now, as I was also perplexed by the unmerged status of the previous one.

daxgames:
bump

daxgames:
Still working on it

Issue #1414, opened 2017-06-22
Change Tasks default start directory


bdr00:
Similar to #203, I want Cmder to start from the opened directory when I run it from Windows explorer address bar, like so:
![cmder_issue](https://user-images.githubusercontent.com/24456233/27443324-5d253118-577b-11e7-95c1-96d6d8529cb4.gif)

I messed a bit with the settings and made it work by removing `%USERPROFILE%` from the task command:
```
cmd /k "%ConEmuDir%\..\init.bat"  -new_console:d:
```
Can this be the new [default](https://github.com/cmderdev/cmder/blob/bee82d00e811875ce4cc90f954f8ca2f6ca44389/config/ConEmu.xml#L491) if it doesn't break anything?

alex-salas:
Being trying to make this work for a couple of days. I didn't have any problems setting this thing up a year ago. I could just go to the address bar and open the terminal in that path 🤔 

Even tho I'm creating a new task with

- ` cmd /k "%ConEmuDir%\..\init.bat" -new_console:%__CD__%`
or
- ` cmd /k "%ConEmuDir%\..\init.bat"  -new_console:d:`

It's not working for me.

![image](https://user-images.githubusercontent.com/29067203/35368344-f16ee4f2-0150-11e8-8e0b-1edccb2baada.png)

![image](https://user-images.githubusercontent.com/29067203/35368360-ff44247a-0150-11e8-9c4e-4ff0b4690bd1.png)

bdr00:
Which version do you have installed? I'll try to reproduce this.

alex-salas:
@bdr00 I have `v1.3.4`

bdr00:
@alex-salas this might not be the best solution, but looks like `%ConEmuWorkDir%` holds the path:
![image](https://user-images.githubusercontent.com/24456233/35522326-a5898e8e-0524-11e8-8391-64b93db56b5d.png)

Try adding startup commands to your batch/powershell user profile files respectively (located in `%CMDER_ROOT%\config\`):
```batch
REM user-profile.cmd
pushd %ConEmuWorkDir%
```
```PowerShell
# user-profile.ps1
push-location -path "$env:ConEmuWorkDir"
```

HTH

alex-salas:
Can confirm this is working after adding the code @bdr00  mentioned in the `user-profile.cmd` file.

[...]

PR #1772, opened 2018-05-10
Set cmderStart to launcher's current working directory as fallback

bnbdr:
When you open the launcher from the address bar windows sets the path as its working directory(instead of passing it as an argument). The launcher doesn't pass that option to ConEmu, which ends up using `%USERPROFILE%`:

image

This is a more elegant fix for the workaround mention in #1414, which stopped working because `%ConEmuWorkDir%` isn't set in the host process.

### DIFF
--- a/launcher/src/CmderLauncher.cpp
+++ b/launcher/src/CmderLauncher.cpp
@@ -433,7 +433,25 @@ cmderOptions GetOption()
 			cmderOptions.error = true;
 		}
 	}
-
+	
+	if (cmderOptions.cmderStart == L"") 
+	{
+		TCHAR currentWorkingDir[MAX_PATH];
+		if (GetCurrentDirectory(MAX_PATH, currentWorkingDir) == 0)
+		{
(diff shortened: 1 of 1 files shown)


bnbdr:
Alright, thanks.

alex-salas:
I installed Docker in windows today and I can't seem to make it work again.
Any new ways to make cmder open from the windows address bar? 🤔

bnbdr:
@alex-salas if you're too lazy to build it yourself you can replace your `Cmder.exe` with my [compiled binary](https://github.com/bnbdr/cmder/releases/tag/fx_start_dir).

*I do not condone running binaries from unverified sources*

daxgames:
@alex-salas and @bnbdr no reason to build it yourself.  The nightly builds are packaged just like the release builds they are just not released.  There is a link at the bottom if the readme.md to the nightly builds page

daxgames:
@bnbdr can we close this since it is no longr necessary any longer?

bnbdr:
@daxgames  Yeah, thanks 👍

Issue #1512, opened 2017-10-11
Ctrl-A move cursor to wrong position


3usi9:
Version: v1.3.2
OS: Windows 10 64-bits

I am used to deleting current commands with ```C-a C-k``` in terminal..
however, it seems doesn't work on cmder...

Input one word:

<img width="53" alt="cap1" src="https://user-images.githubusercontent.com/20506941/31433273-66ca7534-aeab-11e7-80cc-524285dafe5e.PNG">

Everything seems OK..

Input two words:
<img width="61" alt="cap2" src="https://user-images.githubusercontent.com/20506941/31433308-7d91c5ce-aeab-11e7-99dd-b248a3009fd5.PNG">

After enter ```C-a```
<img width="64" alt="cap3" src="https://user-images.githubusercontent.com/20506941/31433333-8f11b34a-aeab-11e7-9d94-aec78e90fa32.PNG">

After ```C-k```
<img width="61" alt="cap4" src="https://user-images.githubusercontent.com/20506941/31433350-9a73c386-aeab-11e7-8569-c2cb74bc17c8.png">

Even if there is no command to be executed, a character is left here...

balalaXD:
Same issues in 
Version:1.3.6
OS: Windows 7 Ultimate x64

Stanzilla:
Hey there,
Cmder uses ConEmu as terminal emulator, if you are experiencing a problem with that, please open a bug over at https://github.com/Maximus5/ConEmu

Thank you! :)

balalaXD:
Open your `%CMDER_PATH%/vendor/clink.lua`,
Change line `local lambda = "λ"`  with  `local lambda = "$"`

Ref: 
#1425 
https://unix.stackexchange.com/questions/207941/custom-bash-prompt-cursor-positioning-issue
https://bixense.com/clicolors/

PR #1790, opened 2018-05-30
Fix wrong position by Ctrl-A

balalaXD:
Reproduce:
With settings of `local lambda = "λ"`, input `12345` then hit `Contrl-A`, you will found cursor located between `1` and `2`

As #1512 describe


Stanzilla:
We're not going to replace the iconic lambda, the real problem is either clink or conemu not treating that symbol correctly

PR #1834, opened 2018-07-06
Refactored the Mercurial prompt code to be more efficient.

b0bh00d:
I refactored the HG prompt to use a much faster Mercurial command that provides all the needed information in a single short line.  The original code was taking up to 10 seconds on my repository.  This code takes about 1.


Stanzilla:
Thanks!

PR #1860, opened 2018-08-23
Add %LANG% support

gucong3000:
- Add support for environment variables `%LANG%`


daxgames:
This will probably break cmder_mini

Stanzilla:
oh, damn

daxgames:
Just guessing I have not tested it.

Stanzilla:
It only runs when git is found so that should be correct

daxgames:
Cool

Issue #1695, opened 2018-03-07
How to store configuration


DRSDavidSoft:
We need to come up with a standardized way to store Cmder configuration [here](https://github.com/cmderdev/cmder/tree/master/config).

I'm working on a branch to customize the prompt color and handle character (currently λ) from the `/config` directory, and similarly, a lot of `init.bat` (e.g. @daxgames's recursive path enhancing) could benefit from a stored configuration file.

@cmderdev/trusted-contributors - What would you vote as a suitable container for the Cmder configuration?
- XML
- YML
- JSON
- TXT/INI
- BAT/SH
- (other)

Keep in mind that it needs to be:
- easy to edit for user
- accessible from both batch and bash environments
- uniform across all vendored platforms

DRSDavidSoft:
@daxgames That's good (and looking forward to your PR) – though by not unified, I mean that each shell is currently loading its custom script. e.g. for changing the prompt color from green to cyan, I have to modify these <u>3</u> files:

- git-prompt.sh
- clink.lua
- profile.ps1

It would be better to have a single `console.lua` file in /config, which then would be parsed by each of those files using Lua executable. Additional _platform-specific_ scripts then would be separated to .sh, .bat, .ps1 and .lua scripts in the `/config` directory.

So, a newbie user could easily edit that `console.lua` file to achieve customizing the prompt, essentially instructing the `/vendor`-based scripts on _how_ to set the prompt. If further modification is required, (besides other unified files such as `git.lua`, `path-enhancing.lua`, etc...) then he/she can write a custom script to run on start-up.

I see _many_ parallels between platforms that Cmder supports, besides customizing prompt. That's what I meant, for example, to set the path depth limit from `/config`. Also, that's why I suggested container formats like XML, YML, and JSON. Because they're specifically made to hold _data_, rather than execute code. (I'm fine with .lua, though, as it's already easy to edit.)

There are [methods](https://gist.github.com/pkuczynski/8665367) for bash to parse .yml, or better we can include a lua.exe binary (<50KB) in the `/vendor` directory to parse console.lua and adjust color and the λ handle accordingly.

[...]

PR #1873, opened 2018-09-05
Launcher improvements

DRSDavidSoft:
#### 1. ✓ Move strings to separate `strings.rc2` resource file
Strings that aren't short will be migrated to [here](https://github.com/DRSDavidSoft/cmder/blob/9b121f1c0aac863c8124443e306cb3d8e67647cc/launcher/src/strings.rc2#L2-L12) instead of being defined in the C++ code like this:
https://github.com/cmderdev/cmder/blob/49da3745bce7c4f850079c0cc5a92ee978207b56/launcher/src/CmderLauncher.cpp#L300-L303
<br />

#### 2. ✓ Switch to `TaskDialog` instead of `MessageBox` for displaying messages
Since XP has been dropped (#1707), we can switch to **[TaskDialog](https://docs.microsoft.com/en-us/windows/desktop/api/commctrl/nf-commctrl-taskdialog)** to deliver a better message window, as discussed in https://github.com/cmderdev/cmder/issues/1726#issuecomment-375939159

taskdialog-cmder-help
<br />

#### 3. Move functions to a class – _TODO_
_TBA_ – General refactoring of the code.
<br />

#### 4. Detect and configure `%CMDER_ROOT%` variable – _TODO_
_TBA_ – When registering the application and the context menu, Cmder will check for and correct `%CMDER_ROOT%` if needed.
<br/>

#### 5. Add switches for `/version` and `/help` – _TODO_
_TBA_ – Useful to quickly check for the current binary version.


daxgames:
I am not a c++ dev so I can't tell you how the code should or should not be organized.  At best I hack at it until I get the result I want, so have at it.

I'm not sure how %cmder_root% could be wrong since it is based on the exe dir of the launcher exe.  Not sure what you mean by discover and configure.

I don't think there needs to be a check for update feature nor an auto update feature.  Its a portable app you download, expand and copy it to the dir, it's just not necessary.  This is probably the best idea of them all, while I think it's unnecessary and could be confusing since conemu already has this and it is often confused with a cmder updater users may like it.

Why would you add a confirmation dialog for a command line switch the user typed, isn't that confirmation enough.

No checkboxes, no registry settings, no more config files.  Btw what is a 'do not as me again box' do?

daxgames:
Actually I'd like the version file to be called `version` and contain the version so that when someone upgrades they don't end up with multiple version files.  A `/h` and a `/v` switch might be nice and easy to add.

DRSDavidSoft:
Thanks for your input, @daxgames.
Here's my explanations:

### Variable detection
> I'm not sure how %cmder_root% could be wrong since it is based on the exe dir of the launcher exe. Not sure what you mean by discover and configure.
There are possible reasons why the content of the variable could be wrong:
- The directory has been moved (e.g. from Downloads to Desktop)
- The mounting partition could change (e.g. the memory stick could change from E: to G:)
- A previously defined variable could interfere (e.g. There's a system variable defined, but we need to set a user variable)

The plan is to check the content of the variable upon startup, and display the following message when an incorrect value is detected.

| Variable warning | Variable warning (more details) |
| --- | --- |
| cmder-variable-warning-1 | cmder-variable-warning-2 |

BTW, one use of 'Do not ask me again' is here (the checkbox in the bottom.)

<br />

### Check for updates
> I don't think there needs to be a check for update feature nor an auto update feature. Its a portable app you download, expand and copy it to the dir, it's just not necessary. This is probably the best idea of them all, while I think it's unnecessary and could be confusing since conemu already has this and it is often confused with a cmder updater users may like it.

I understand, however I do still suggest a behaviour that matches the portable downloaded apps aspect. Just running a simple check for updates (and linking users to the update page where they can download themselves.) Simply put, if there is an update on GitHub releases, a simple message could be shown, asking the use if they'd like to open the download page.

cmder-update-concept

As for the **ConEmu feature**, I think its `Update.CheckOnStartup` should be set to disabled in Cmder if it's not already. This will help prevent confusion.
The plan is to write a good update routine in the Cmder Launcher itself that takes care of compatibility check, configuration backups, and confirmations / notifications for the user.

<br />

### Command-line confirmation

> Why would you add a confirmation dialog for a command line switch the user typed, isn't that confirmation enough.

Hmm, you right about the user typed switches already being confirmations. The idea was to use the `/register` and `/unregister` in the unofficial setup scripts (i.e. _'Uninstall'_ would call `Cmder.exe /unregister) which would ask the user if they wanted to remove Cmder.

I think my thinking was not valid. Those confirmation pages should go in those scripts directly, and calling any switches should perform the action silently.

### How to store settings

> Btw what does a 'do not as me again box' do? Is it for the proposed features like check for update?

Any dialog where the user may choose an action 'as always', including check for updates.

Now, a follow up question: If I implement this, should I define separate variables for each routine (e.g. downloads, checks, etc), or should they all follow a single 'do not ask me again' directive?

<br />

> No checkboxes, no registry settings, no more config files. [...] If you are storing additional config it should be in a file in %cmder_root%/config

Ouch! I mean I still agree that the location of the config should be in `%cmder_root%/config`, but I'm not sure on _how_ to store the individual settings for the launcher. Do I use `launcher.txt`? `launcher.ini`? `launcher.xml`? (See https://github.com/cmderdev/cmder/issues/1695#issue-303264076)

Just a thought though, if we are already modifying the Registry settings for the context menu, IMHO it means that it's very likely that the user is using Cmder in non-portable mode.

<br />

### The '`version-*`' file(s)
> Actually I'd like the version file to be called `version` and contain the version instead of `version-1.3.6.675` so that when someone upgrades they don't end up with multiple version files

👍 on the idea. I also dislike having multiple `version-*` files in the root directory after an update.

However, actually the file's name is being read by the user to check for the version of Cmder, not its content in a text reader. I'd vote to use #1729's method instead of opening the file with an editor, which is easier and more consistent.

I'd suggest either:

1. Getting rid of the file completely and use Cmd.exe's properties to check for version instead
2. Writing a simple routine to remove older `version-*` files in the root directory

DRSDavidSoft:
@daxgames It's okay, I actually like your counter arguments.
<br/>

> The examples talk about the `%CMDER_ROOT%` variable which I still don't see how it could be wrong

IMO, The one use case (moving it around when the context menu is registered) justifies the reason. 
However, if you don't see how it could be wrong, neither the user will. If the variable is really correct, the code will simply ignore and pass the warning; but if the variable is either _changed_ or _set incorrectly_, a simple warning will pop up asking the user if they want Cmder to fix it. I see no UX harm in that.
<br />

> It could become VERY annoying though to someone like me that has multiple copies of `cmder` installed for testing purposes.

Solution: Do not ask me again :) (either ticking the checkbox, or a simple `.ini` line, whichever is the simplest).

I have multiple copies of Cmder as well (I think for each branch that I'm working on, I have one)

So once (or if) this PR gets merged, the annoyance will be only be setting that 'Do not ask me again' flag _once_.
<br/>

> Maybe I'm old school but xml is probably overkill to store a set of binary flags to save checkbox state. Ini file would probably suffice and I would do a separate flag for each checkbox.

I agree with you, XML _is_ actually overkill just to store some flags. I also prefer a `.ini` file to store them. (The `.ini` files also go really well with Win32 c++ programs!)
The main concern was scalability, if the launcher was ever to incorporate more features some day, but I'm already thinking against that.
<br/>

> I prefer file storage to registry storage because at its core cmder is a portable app. 

Couldn't agree more. Portability is the #<i></i>1 feature of Cmder.

<hr />

### Now about the maintainability discussion

> My biggest concern with your ideas is maintainability. Seems like a lot to add to an otherwise simple launcher. We barely support the launcher today because none of us are really c++ experts. You may know it but what if you get bored with the exciting world of contributing to cmder and decide your done.

> I have said before I am concerned with adding so much functionality to the launcher that it becomes difficult to maintain.

I **completely** understand your point, I wouldn't want having a lot of unmanageable code added to my repo either. However, I should say that this launcher wouldn't have been possible if @austinwagner didn't made [a PR](https://github.com/cmderdev/cmder/pull/62), or if he did and @samvasko didn't approve it. Many contributors to any project come and go, and they build upon the project as they come.

I don't consider myself a C++ expert. As you said, I just know how to do it. But it's against my nature to toy with something and get bored with it. I might stop contributing one day, but I won't leave a ton of unfinished, unmaintained code in the repo when I do it. Furthermore, that day _isn't_ today. I can assure you I'll have interest in maintaining the features that I introduce for many more days to come.

The only issue that remains IMHO, is to decide what features to add, and which ones to drop. I won't introduce any unnecessary changes into the launcher code.

So, as long as we sort out the inevitable issues that arise with each new feature, I would ask that to be added to `Cmder`.

DRSDavidSoft:
↩️

MartiUK:
@DRSDavidSoft is this still WIP and unable to merge?

DRSDavidSoft:
@MartiUK I'm sorry, but I have committed a lot of changes that need to be pushed before this PR is ready to be merged.

I haven't had time to polish and test my changes (in addition to resolving a couple of conflicts), so please keep this on hold until I can complete everything.

I'll post an update when the changes are ready to be pushed. Thanks for the patience!

DRSDavidSoft:
I'm temporarily closing this PR as I've been busy in the past couple of months, and don't want the bot to close it as stale.

Once I finish working on it, I will reopen the PR.

MartiUK:
@DRSDavidSoft Okay mate, no rush

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #1938, opened 2018-11-02
Fix documentation

JosefPihrt:
* Fix typos
* Fix broken link
* Remove escaping inside inline code
* Replace single quote with backtick where inline code was intended


Stanzilla:
Thank you!

JosefPihrt:
You're welcome.

PR #1918, opened 2018-10-20
add diag helper scripts and adds to the path

daxgames:
These scripts are to help us help our users by getting a relative picture of their environment that they can submit with an incident.

PR #1940, opened 2018-11-03
Initialize time_init to fix init error message

glucas:
Fix to #1918: Make sure %time_init% is initialized, otherwise this line causes an error each time a cmd shell is created:

    if %time_init% gtr 0


Stanzilla:
Thanks!

daxgames:
@glucas Thank you.  I never turned it off after writing this.

PR #1969, opened 2018-11-30
Improve LANG env var detection

### LINKED ISSUE #1956: Git detection should use env from git install root
### Purpose of the issue
- [x] Bug report (encountered problems/errors)
- [ ] Question

### Description of the issue 

Version: Cmder 1.3.8, installed through chocolatey.

When I start a new shell in Cmder I get the following line before the prompt.
```
env: /usr/bin/locale: No such file or directory

E:\Dev                                           
λ 
```
I believe this comes from init.bat line 247:  `for /F "delims=" %%F in ('env /usr/bin/locale -uU 2') do (` and it is probably caused by me having an env.exe in my path before the one from git.
I changed it locally to `for /F "delims=" %%F in ('"%GIT_INSTALL_ROOT%\usr\bin\env" /usr/bin/locale -uU 2') do (` which works for me.

My output for `where env` and git from a plain command prompt:
```
e:\Dev>where env
C:\GnuWin\bin\env.exe

e:\Dev>where git
C:\Program Files\Git\cmd\git.exe
```

Doing same from a cmder prompt:
```
E:\Dev
λ where env
C:\GnuWin\bin\env.exe
C:\Program Files\Git\usr\bin\env.exe

E:\Dev
λ where git
C:\Program Files\Git\cmd\git.exe
C:\Program Files\Git\mingw64\bin\git.exe
```

Note: GnuWin 0.6.3.1, Git 2.19.1 and Cmder 1.3.8 were all installed through Chocolatey.

rjvdboon:
Change the detection of the LANG env var:
- only try to detect if it does not set yet
- search for a proper locale.exe
-- prefer locale.exe from the git install root
-- check the path for a locale.exe
-- use git installed env.exe with /usr/bin/locale`
-- fallback to the old env /usr/bin/locale

Should we actually always set the LANG env var, even if the detection fails? This is what they did in [Hyper](https://github.com/zeit/hyper/issues/346).

Fixes #1956

### DIFF
--- a/vendor/init.bat
+++ b/vendor/init.bat
@@ -243,9 +243,16 @@ if defined GIT_INSTALL_ROOT (
 
     :: define SVN_SSH so we can use git svn with ssh svn repositories
     if not defined SVN_SSH set "SVN_SSH=%GIT_INSTALL_ROOT:\=\\%\\bin\\ssh.exe"
-
-    for /F "delims=" %%F in ('env /usr/bin/locale -uU 2') do (
-        set "LANG=%%F"
+    
+    if not defined LANG (
+        :: Find locale.exe: From the git install root, from the path, using the git installed env, or fallback using the env from the path.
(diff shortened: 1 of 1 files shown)


daxgames:
@rjvdboon Thanks for the PR.

PR #1975, opened 2018-12-03
Add /git_no_unix switch to init.bat to prevent adding !GIT_INSTALL_ROOT!\usr\bin to PATH

patstew:
Adding all the unix tools to the path prevents some programs (Vivado) from working for me.

### DIFF
--- a/vendor/init.bat
+++ b/vendor/init.bat
@@ -14,6 +14,8 @@ set debug_output=0
 set time_init=0
 set fast_init=0
 set max_depth=1
+:: Add *nix tools to end of path. 0 turns off *nix tools.
+set nix_tools=1
 set "CMDER_USER_FLAGS= "
 
 :: Find root dir
@@ -78,6 +80,16 @@ call "%cmder_root%\vendor\lib\lib_profile"
         ) else (
             %lib_console% show_error "The Git install root folder "%~2", you specified does not exist!"
             exit /b
+        )
+    ) else if /i "%1"=="/nix_tools" (
+        if "%2" equ "0" (
+            REM Do not add *nix tools to path
+            set nix_tools=0
+            shift
+        ) else if "%2" equ "1" (
+            REM Add *nix tools to end of path
+            set nix_tools=1
+            shift
         )
     ) else if /i "%1" == "/home" (
         if exist "%~2" (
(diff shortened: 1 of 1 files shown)


daxgames:
I'm not sure about this feature.  If you don't want *nix tools why not use `cmder_mini`.

daxgames:
or remove the unix tools from the path using youe `user_profile.cmd`

patstew:
I am using cmder_mini, but cmder still picks up my git for windows installation. I like having the git prompt integration etc, and want 'git' on the path, but having 'sh' _anywhere_ on the path causes problems for me. I'd rather just maintain my own init.bat with that line commented than write a script to edit the PATH afterwards. Git for windows offers 2 options for path integration when you install, with and without the unix tools, I assume because they know it can be a problem (not just one that affects me).

daxgames:
I understand.  I am working on a switch to `/prefer_nix` that adds `*nix` commands to the front of the path vs. the default of adding to the end of the path.  This is how `Cmder` used to setup the path and users have requested `*nix` tool preference.

FWIW you can do this today with existing conditional call features in Cmder and `init.bat`:  

1. Clone a Conemu Cmder task, call it `Cmder No Unix Tools`
2. Add `/no_nix` to the command line.
3. Create a `%cmder_root%\bin\no_nix.bat` file with:

```
set path=%PATH:C:\Users\user\cmderdev\vendor\git-for-windows\usr\bin;=%
```

4. Edit `%cmder_root%\config\user_profile.cmd` and add.
```
%ccall% /no_nix no_nix
```

daxgames:
@patstew A clarification.  I just want to make sure your change works with mine before merging.

vendor/init.bat
```diff
@@ -239,7 +242,9 @@ if defined GIT_INSTALL_ROOT (
     ) else if exist "!GIT_INSTALL_ROOT!\mingw64" (
         %lib_path% enhance_path "!GIT_INSTALL_ROOT!\mingw64\bin" append
     )
-    %lib_path% enhance_path "!GIT_INSTALL_ROOT!\usr\bin" append
+    if %git_unix_path% gtr 0 (
```
daxgames: Change to:

```
if "%nix_tools%" == "0" (
```

vendor/init.bat
```diff
@@ -79,6 +80,8 @@ call "%cmder_root%\vendor\lib\lib_profile"
             %lib_console% show_error "The Git install root folder "%~2", you specified does not exist!"
             exit /b
         )
+    ) else if /i "%1"=="/git_no_unix" (
+        set git_unix_path=0
```
daxgames: Delete line 84

daxgames:
👍 LGTM

daxgames:
@Stanzilla - You want to look.  I tested it and it works.

Stanzilla:
LGTM!

daxgames:
@patstew you have any issues with #2222 that implements `/nix_tools 0` should prevent adding !GIT_INSTALL_ROOT!\mingw64\bin to PATH #2214

PR #1982, opened 2018-12-08
make /register work with /single

haroldhues:
Similar to 787ac3773cab86d31f477abd1a31808a7c39dfe5 it is handy to be able to make the `/single` part of the shell menu command too.


daxgames:
LGTM 👍 @haroldhues Thanks for the PR!

Issue #1985, opened 2018-12-13
VSCode tasks don't run properly on Cmder


eparizzi:
I've followed the instructions [here](https://github.com/cmderdev/cmder/wiki/Seamless-VS-Code-Integration) and I'm able to run Cmder as VSCode integrated terminal.
Everything seems to be working fine, except for VSCode > Run Task which ouputs a lot of stuff but doesn't run the task.

Example:
VSCode > Terminal > Run Task > npm start

Integrated terminals shows:
```
Executing task: npm start
DEBUG(init.bat): Env Var - CMDER_ROOT=D:\Portables\cmder

DEBUG(init.bat): Env Var - debug-output=1

DEBUG(init.bat): Looking for Git install root...

DEBUG(:read_version): Env Var - git_executable=D:\Portables\cmder\vendor\git-for-windows\cmd\git.exe

DEBUG(:read_version): D:\Portables\cmder\vendor\git-for-windows\cmd\git.exe does not exist.

DEBUG(:validate_version): Found Git Version for VENDORED: ...

DEBUG(:read_version): Env Var - git_executable=C:\Program Files\Git\cmd\git.exe

DEBUG(:read_version): Env Var - GIT_VERSION_USER=2.18.0.windows.1

DEBUG(:validate_version): Found Git Version for USER: ...

DEBUG(:enhance_path): Env Var - find_query=C:\\Program\ Files\\Git\\cmd

DEBUG(:enhance_path): Env Var 1 - found=0

DEBUG(:enhance_path): Env Var 2 - found=1

DEBUG(:enhance_path): Env Var - find_query=C:\\Program\ Files\\Git\\mingw64

DEBUG(:enhance_path): Env Var 1 - found=0

DEBUG(:enhance_path): Env Var 2 - found=0

[...]

PR #2004, opened 2019-01-06
Workaround vscode tasks showing cmder debug output

daxgames:
Not sure if this is the right solution to #1985 but it works for me.  Someone please confirm.


DRSDavidSoft:
Not the best solution to the problem

daxgames:
@DRSDavidSoft Awesome, hopefully you have a better idea.  I toyed with just skipping the `/d` if a specific vscode env var was present(the first commit) but that seemed hacky too.

PR #2013, opened 2019-01-17
Parsing decodeURI

SCWR:
Don't be uris allow characters will be encoded,hence the %% case.
The encoded string is parsed through the decodeURI function

### DIFF
--- a/vendor/clink.lua
+++ b/vendor/clink.lua
@@ -17,8 +17,8 @@ dofile(clink_lua_file)
 -- Makes a string safe to use as the replacement in string.gsub
 ---
 local function verbatim(s)
-    s = string.gsub(s, "%%", "%%%%")
-    return s
+--     s = string.gsub(s, "%%", "%%%%")
+    return decodeURI(s)
 end
 
 ---
@@ -413,4 +413,14 @@ if clink.get_env('CMDER_USER_CONFIG') then
     -- so config reloading using Alt-Q won't reload updated modules.
     dofile(filename)
   end
+end
+
+function decodeURI(s)
+    s = string.gsub(s, '%%(%x%x)', function(h) return string.char(tonumber(h, 16)) end)
+    return s
+end
+
+function encodeURI(s)
+    s = string.gsub(s, "([^%w%.%- ])", function(c) return string.format("%%%02X", string.byte(c)) end)
+    return string.gsub(s, " ", "+")
 end


Stanzilla:
Can you explain in more detail which problem you are fixing?

SCWR:
First cmder is really easy to use, so I used it on vscode.
Remember that the version used for the first time is 1.3.5.
The version control tools used by my company project are git and svn. When I use it under git, everything works fine, but one day I developed a project under svn, and the vscode terminal crashed. I checked the lua script error. Over.
Because lua is not a development language I am proficient in, so I try to google it. I found it to be
```
String.gsub(clink.prompt.value, "{svn}", color.."("..branch..")")
```
Because Chinese is encoded after %% of the situation, and then in the string replacement may be judged as an escape code resulting in an error.
Why is there no problem under git, I think that the people who use git are basically programmers, and because the existence of github and the recommended use of English, the project path is generally English.
The svn situation is more complicated. At the beginning, the person in charge started to build a Chinese project, such as 代码管理, and then used English to build the project, so there is a Chinese path. ╮(╯▽╰)╭
So the way I solved it for the first time is the same as this one.
```
Branch = string.gsub(branch, "%%", "%%%%")
```
The program did not report an error, but the output text is as follows.
snipaste_2019-01-18_09-33-09
I have a little obsessive-compulsive disorder, I don't think it is very good. I found that because Chinese is encoded, can I transcode it? So I google it again. Get this plan.
The result was successful like this.
snipaste_2019-01-18_09-38-16
After that, every time I upgrade the cmder, I need to change the change to the new version of the lua script compared to the previous lua script.
Until this upgrade, I found that the solution was consistent with my first solution. So I boldly fork it, modify and propose the merge application, so I don't need to manually modify the lua script for the next upgrade. As I said before, lua is not a development language that I am proficient in, so I hope you can check it out and make the correct changes.

When I submitted the merger application, I thought that there will be a %% situation after decoding.
So I feel like this change.
```
Local function verbatim(s)
    Return string.gsub(decodeURI(s), "%%", "%%%%")
End
```
I hope to help you a little.

Finally, cmder is really easy to use, ^_^

daxgames:
Can someone that uses SVN confirm this does not break anything?

chrisant996:
Merging this would make the prompt show inaccurate path names, for example if a path name is literally "My%20Home" (as was recently mentioned in #2859) then it would be displayed as "My Home" which is inaccurate (typing `cd "My Home"` would fail because that's not the actual path name in the file system).

Svn should only convert Unicode URIs into URI-escaped strings (e.g. "https://some.where/over%20there").  It should not convert Unicode path names into URI-escaped strings.  URIs and path names do not have the same escaping rules.  If svn is performing that conversion automatically when writing output, then that seems like a bug or a misconfiguration.

If someone manually converted Unicode path names in the file system to use URI escaping rules so that the path names in the file system have "%XX" sequences in them, then Cmder prompt should not translate those -- that would be misleading.  It would be reasonable for a user or a team to make a custom prompt filter that performs such translations, but it shouldn't be the default all users.

Also, the `verbatim()` function already has a specific different purpose.  Even if URI decoding were to be applied for all users, it should be done in a separate function, not in the `verbatim()` function.  And also, PR #2861 removes the `verbatim()` function and adds a `gsub_plain()` function instead, to simplify and clarify what the Cmder prompt string find/replace functions are trying to do.

DRSDavidSoft:
Closing this, as `verbatim()` was not the correct function to implement this, and it shouldn't be enabled by default anyway, and #2861 was merged, which this PR conflicts with that.

@chrisant996 Can you please write a short and working prompt filter that would call `decodeURI()` on a part or the whole of prompt? I believe it would serve as a nice example for the future Cmder users who might want to implement features like this for their prompt.

If it's too much trouble, I'd also appreciate a very short guide or writeup, or even pointing to the relevant section of the documentation for Clink.

Thanks!

chrisant996:
Cmder already includes the scripts from clink-completions under `vendor\clink-completions`, and the [git_prompt.lua](https://github.com/vladimir-kotikov/clink-completions/blob/master/git_prompt.lua) script demonstrates how to do post-processing to modify prompt text that's been generated by a different prompt filter.

[Clink docs](https://chrisant996.github.io/clink/clink.html) have quite a lot of content about writing prompt filters, and the section [Customizing the Prompt](https://chrisant996.github.io/clink/clink.html#customizing-the-prompt) is a good place to start.

I don't think decodeURI is a good example to create, as it's only applicable to maybe a handful of people who have done very unusual (and problematic!) manual encoding of paths in svn, seemingly to work around not knowing how to configure svn properly.  The git_prompt.lua example is much better and much more widely applicable.

PR #2060, opened 2019-03-16
Fixes and More Cmder Like Powershell

### LINKED ISSUE #2018: oh-my-posh startup scripts in user_profile.ps1 don't work
### Purpose of the issue
- [x] Bug report (encountered problems/errors)
- [ ] Question

### Description of the issue 
Put oh-my-posh startup scripts in user_profile.ps1:
```
Import-Module posh-git
Import-Module oh-my-posh
Set-Theme Paradox
```
It doesn't work. If these scripts are written to profile.ps1, it works.
Is there something wrong?

daxgames:
# Fixes

## Git detection in `Cmd::Cmder` Sessions

* Fix setting env variable for found versions.

## PS1 Prompt Customization -  #2018
   * Moves Cmder prompt settings after running `config\user_profile.ps1` and `config\profile.d\*.ps1` scripts.
   * Only sets Cmder prompt if it is currently set to the Powershell default value.  `$PostPrompt` kinda worked but not completely and it is still there.

# Adds 

## More Cmder Like

* Powershell sessions no longer assume Git is available.  Works better with `Cmder Mini`
* Powershell sessions using Cmder embedded Git also have `%git_install_root%\usr\bin` and `%git_install_root%\mingw[32|64]\bin` added to the path.
* `user_profile.ps1.default` provides an easy way, by un-commenting, to delete the following default Powershell aliases that conflict with `Bash` commands:

_Note: This is done in User Profile so the user can control it.  Existing users will need to add these settings if they want the functionality_ 

```
cat
clear
cp
diff
echo
kill
ls
mv
ps
pwd
rm
sleep
tee
```

## Add `Cmder Here` to drive roots for `Cmder.exe`


vendor/user_profile.ps1.default
```diff
@@ -23,3 +23,20 @@
 
 ## <Continue to add your own>
 
+# Delete default powershell aliases that conflict with bash commands
+# If you prefer the powershell aliases remove this.
+if (get-command git) {
```
Stanzilla: should we really do that by default?

daxgames: It's only on first run and the user can turn it off if they want.  I feel like anyone that types `curl` in a shell expects to be using `curl` not `invoke-webrequest`.  I know it posses me off.  :-)

We could comment themand let the user uncomment them.

Stanzilla: Hrm yeah I know the problem, just not too sure about changing defaults for users

daxgames: rectified

PR #2099, opened 2019-05-16
Update README.md

warthurton:
Added /t switch for timed init mode.


daxgames:
@warthurton thanks for the pr

PR #2153, opened 2019-08-29
Disable clink logging

mcovalt:
Not much to this. Clink creates a log file in `%LOCALAPPDATA%/clink` by default. This PR adds the argument `--nolog` when launching clink.


daxgames:
Thanks for the PR.  Merged.

PR #2177, opened 2019-09-26
Clink Completions 0.3.4

daxgames:


Stanzilla:
got confused for a sec and thought you really forked clink :D Maybe rename that to say that you updated clink-completions!

PR #2187, opened 2019-10-10
Improvement for user-prompt load in PS

NoNaNy:
Checking if the prompt variables are null o string empty (defined) for don't overwrite them.
Moved the lambda symbol to CmdPrompt for really to user let overwrite. And removed the double space after this symbol.


daxgames:
Resolved merge conflict caused by merging #2193

daxgames:
@NoNaNy All this really does at this time is makes changing the `λ` to another character.  #2193 fixed the Cmder prompt components in a better way so it was merged first.  Please test and make sure it still works for you as you intended.

daxgames:
@NoNaNy and thanks for the PR!

NoNaNy:
Yes this issue was fixed with the PR #2193

Issue #2176, opened 2019-09-26
The issues template has issues.  


daxgames:
### Purpose of the issue
- [ ] Bug report (encountered problems/errors)
- [ ] Feature request (request for new functionality)
- [ ] Question

### Version Information

* Cmder Version:
* ConEmu Version:
* Windows Version(e.g. Windows 10 Pro 1809):

### Description of the issue

It is not clear  where a user should enter the description.  It says:

\<\!\-\- Provide a clear, simple description of your issue here \-\-\>

and users do exactly that.  they edit the comment so the question never shows in the Github interface.  The same is true for other areas of the template.  I can fix it if someone gives me access to do so OR I have edited it here and hopefully make it clearer for future users.

daxgames:
Whoever has access can get the fixed template by editing my above comment  and copying down to and including:

\<\!\-\- Provide a clear, simple description of your issue here \-\-\>

Stanzilla:
I can already tell you that it's too long and nobody is going to read it :( tried longer templates with a few projects before and it never worked out

daxgames:
Agree.  That is the CURRENT template.  I changed only 5 lines

daxgames:
Did not know that is where it was.

Stanzilla:
oh. lol.

DRSDavidSoft:
Apparently the current issue template format is also considered legacy now.
If you do a re-write, you could use [the new format](https://help.github.com/en/articles/about-issue-and-pull-request-templates) and split it into separate files.

PR #2189, opened 2019-10-14
Update template issue

DRSDavidSoft:
This is a temporary fix for #2176 until @daxgames write a better issue template.

### Problem: people write their text in place of "here", inside the comments
```
<!-- Provide a clear, simple description of your issue here -->
```

### Solution: Ask the users to write the line below the comment
```
<!-- Provide a clear, simple description of your issue below this line -->
```

I feel like I should've written the instructions like this from the beginning.  I don't know why I assumed that everyone who posts an issue here would be automatically familiar with HTML comment tags and/or the markdown syntax; since this clearly is not the case. 😄 

Hopefully, this will be a simple yet effective solution.

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -107,7 +107,7 @@
 ### Version Information
 <!-- Please write your Cmder and ConEmu version below this line -->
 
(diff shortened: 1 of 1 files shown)


daxgames:
Hah thats exactly what i was proposing.  I seriously doubt ill rewrite it completely

PR #2335, opened 2020-06-24
Added support for setting custom icons for Cmder window

kvittokonito:
# Summary
Cmder's launcher currently overrides ConEmu's custom icon feature and defaults to using `cmder.ico`.
As a daily user of Cmder that uses a lot of different tasks, it can be difficult to identify which window is running which task in the taskbar.

This PR solves this issue by adding the `/icon <path>` command line argument.  
Examples of how it looks with custom icons:  
* Using `cmder_red.ico` for `{cmd::Cmder as Admin}`:  
  * Window:
Window  
  * Task Bar:
Task Bar  
* Using `cmder_blue.ico` for `{PowerShell:PowerShell}` task:  
  * Window:
Window  
  * Task Bar
Task Bar

## Adds
* Add `/icon <path>` command line argument for choosing a custom icon.
* Add Debug and Release build configurations for x64.

### DIFF
--- a/launcher/CmderLauncher.sln
+++ b/launcher/CmderLauncher.sln
@@ -9,12 +9,18 @@ Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Win32 = Debug|Win32
 		Release|Win32 = Release|Win32
+		Debug|x64 = Debug|x64
+		Release|x64 = Release|x64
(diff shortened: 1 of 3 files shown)


xinghe5:
unsubscribe

daxgames:
@kvittokonito Merged - Thanks for the PR

Issue #2373, opened 2020-08-13
Inconsistent `path` between shell types causing inconsistent Git version and tools if Git is installed locally and a different version than embeded Git


daxgames:
### Purpose of the issue
- [X] Bug report (encountered problems/errors)
- [ ] Feature request (request for new functionality)
- [ ] Question

### Version Information

Cmder Version: 1.3.16

Windows Version:  10 2004 Build 19041.388
Locally installed Git: 2.27.0.windows.1
Embedded Git Version: 2.28.0.windows.1
### Description of the issue

Inconsistent `path` between shell types causing inconsistent Git version and tools if Git is installed locally and a different version than embeded Git

1. Git is installed locally in `C:\Program Files\Git`.
2. Launching `cmd::Cmder` and `powershell::powershell` shells result in inconsistent `path` between shell types causing inconsistent Unix tools availability.
3. If Git is installed locally and a different version than Cmder embedded Git then Git version is also inconsistent.

See screenshot of path from `cmd::cmder`, `powershell::powershell`, `bash::bash` split by delimiter into separate lines:

![image](https://user-images.githubusercontent.com/7318053/90133605-c40ab100-dd3d-11ea-9a97-0aeca22ede80.png)

Result of inconsistencies in each shell:

- `cmd::cmder`
   - Git Version: 2.28.0.windows.1
   - Other Unix tools are available for use in the path.
- `powershell::powershell`
  - Git Version: 2.27.0.windows.1
   - Other Unix tools are NOT available for use in the path.
- `bash::bash` Git Version: 2.27.0.windows.1
   - Git Version: 2.27.0.windows.1
   - Other Unix tools are available for use in the path.

Proposed Fix:

- Use the newer Git version in all shells.
- If Git is installed and configured locally and embedded Git is older.
  - Today we add Unix tools from the local install to the path for `cmd::cmder` but not `powershell::powershell`
  - Proposal:
    - Use the current path and make no changes regarding Git and Unix tools only add Cmder folders.
- If Git is NOT installed and configured locally and we have embedded Git.
  - Today we add `%cmder_root%\vendor\git-for-windows\cmd` to the front of the path
     and  add `%cmder_root%\vendor\git-for-windows\mingw64\bin` and `%cmder_root%\vendor\git-for-windows\usr\bin` after `C:\Windows` to the end of the path
     using embedded Git version and preferring Windows tools.
- If Git is installed and configured locally and embedded Git is newer
[...]

PR #2381, opened 2020-08-29
Powershell Git Version Discovery

daxgames:
This change affects Powershell sessions.

- Provides more consistent path with `cmd::cmder` sessions.(This will also change to match the behavior of this PR in a future PR)
- Allows a user to configure how Cmder modifies the path if a local Git is installed.

See #2373 for the full proposal.

- Find user installed Git on Path
  - If found
    - if newer than Cmder embedded Git
      - Use it's existing Path config and completely ignore embedded Git.
    - Else if Cmder embedded Git exists and is newer
      - Match User installed Git path config using Cmder embedded Git folders.
  - Else if Cmder embedded Git exists 
    - Add Cmder embedded Git folders to the path.
      - `$env:cmder_root\vendor\git-for-windows\cmd;$env:path`
      - `$env:path;$env:cmder_root\vendor\git-for-windows\usr\bin`
      - `$env:path;$env:cmder_root\vendor\git-for-windows\mingw64\bin`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,16 @@
 # Change Log
 
-## 1.3.16-pre
+## [1.3.16](https://github.com/cmderdev/cmder/tree/v1.3.16) (2020-07-29)
 
-* fix vscode terminal errors when setting cmder fast_init=1 outside cmder [#2339](https://github.com/cmderdev/cmder/pull/2339)
+### Fixes
 
+* Merge pull request #2357 from FloSchwalm/fix-git-version-comparison  [Dax T Games]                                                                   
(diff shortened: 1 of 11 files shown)


DRSDavidSoft:
Nice, @daxgames thank you for your continued efforts to improve consistency between Cmd and Powershell.  I especially appreciate the new Git version detection in PS!

daxgames:
@DRSDavidSoft glad you like it.  The cmd.exe side of things is proving difficult so not sure I will be able to implement the full proposal.  I think a MAJOR refactor of the entire lib_git and possibly other Cmder libs may be required.

gancena (approved):

daxgames:
I will be replacing this with a different PR.

PR #2412, opened 2020-11-01
Fix path bug in configured git.

dreamkhv:


daxgames:
What bug is this fixing.  Looks like it breaks more than it fixes by removing a folder from thw path that we want in the path.

MrTraan:
Allright so I'm having issues with this also...
This pr does not fix this issue, but the intention is correct: %GIT_INSTALL_ROOT% evaluates to `"C:\cmder\vendor\git-for-windows\mingw64\bin"` for me.
So "%GIT_INSTALL_ROOT%\usr\bin" evaluates to `"C:\cmder\vendor\git-for-windows\mingw64\bin\usr\bin"` which is not a existing folder (I guess it should be `"C:\cmder\vendor\git-for-windows\usr\bin`)
Because of that I did not have access to unix commands anymore (like ls for example)
I dirty fixed it by setting the same line than the one changed in this commit by
` %lib_path% enhance_path "%GIT_INSTALL_ROOT%\..\..\usr\bin" %path_position%`

daxgames:
@MrTraan what version and type, mini or full, of cmder?  I have never seen it evaluate `%GIT_INSTALL_ROOT%` to that path.

MrTraan:
I'm running the latest (v1.3.16 from the website) full version. I honestly have no idea what happened because it used to work fine and it just broke and I can't figure out what happened.
The only thing that might had an impact is that I installed some packages use chocolatey in the meantime (vim and python).

PR #2430, opened 2020-12-07
Improve init speed

qfp64:
Hello,

The init scripts (`init.bat` and the files in `vendor/lib`) are full of calls to `lib_console` for debug/verbose logging. Unfortunately these calls have a cost which adds up to quite a lot. The reason is that doing e.g. `%lib_console% debug_output ....` invokes `call`, which is relatively slow. This cost is paid even if debug/verbose logging is turned off, which of course it is most of the time.

I have replaced such lines with `%print_debug%`/`%print_verbose%` instead, which first check that the relevant flag is set. I have added `%print_error%` for consistency, though this doesn't give any performance improvement. The original way to print, `%lib_console% debug_output ...` still works in case it's used somewhere else.

On my machine, without this change, the latest version of cmder_mini takes 1.41 sec on average to init (measured with time_init flag). With the change it takes 1.02 sec, so almost a 30% improvement.

### DIFF
--- a/vendor/init.bat
+++ b/vendor/init.bat
@@ -67,7 +67,7 @@ call "%cmder_root%\vendor\lib\lib_profile"
             set "max_depth=%~2"
             shift
         ) else (
-            %lib_console% show_error "'/max_depth' requires a number between 1 and 5!"
+            %print_error% "'/max_depth' requires a number between 1 and 5!"
             exit /b
         )
     ) else if /i "%1" == "/c" (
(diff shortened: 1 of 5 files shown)


daxgames:
I will take a look.  I already have a branch I am working on the gets rid of everything in `%cmder_root%\vendor\lib`

daxgames:
I tested the PR and it did not work as expected.  I sped things up but I got:

image

Also enabling debug does not show any debug messages correctly.

image

vs.

image

qfp64:
Oops, that last commit should fix things. The "if" errors were coming from the delayed expansion "!print_debug!" calls. I've replaced them with %print_debug% as I can't see why the expansion would change - hopefully you can confirm that's ok.

daxgames:
Results of further testing:

Yours : Master : Mine

image

daxgames:
Thanks for the PR - Merged

qfp64:
Thanks for maintaining cmder! I look forward to your even faster version.

DRSDavidSoft:
Unfortunately, this is having a side effect:

If the `%debug_output%`/`%verbose_output%` variables are _not_ set before `init.bat` is ran, `if %debug_output% gtr 0` is always hardcoded to `if 0 gtr 0` _before_ those variables are set in the `:var_loop` loop.

PR #2440, opened 2020-12-16
Update clink-completions to 0.3.6

vladimir-kotikov:
https://github.com/vladimir-kotikov/clink-completions/releases/tag/0.3.6


MartiUK:
Thanks Vladimir

PR #2454, opened 2021-01-01
Upgrade to latest version of ConEmu

naahiv:
Fixes some important issues, specifically pertaining to users using their own versions of Git and vim, for example, which ConEmu fixed in its October release.


daxgames:
Replaced by #2452

PR #2464, opened 2021-01-16
Fix: "The system cannot find the file specified" if %GIT_INSTALL_ROOT%\usr\bin doesn't exist

stephensolis:
I have Git in my path but not through git-for-windows, I have compiled it myself. Therefore my Git isn't in a `usr/bin` structure, it's just in a folder by itself. When I run Cmder I get an annoying message on startup:

```
The system cannot find the file specified.
```

This adds a check so this message will not appear.


daxgames:
Thanks for the PR - Merged

PR #2475, opened 2021-02-08
Fix some commands like 'ls' are not available

simaek:


daxgames:
@simaek Thanks for the PR, merged

Issue #2484, opened 2021-02-25
Really slow prompt inside large git repositories even after disabling git status. 


itsagam:
### Purpose of the issue
- [ ] Bug report (encountered problems/errors)

### Version Information
Cmder v1.3.17.1082, ConEmu v191012

### Description of the issue
Even after you disable git status with the instructions provided in the README file, prompts appear after much delay inside large git repos. Deleting `vendor\clink-completions\git_prompt.lua` does the trick, however.

chrisant996:
Apparently Cmder has _two_ git prompt filters.

The one inside `vendor\clink.lua` checks the `cmder.status` and `cmder.cmdstatus` git config settings.

The one inside `vendor\clink-completions\git_prompt.lua` does not.

Having two separate git prompt filters (and mercurial, and etc) seems like it might additionally be doubling the delay by doing everything twice.

daxgames:
So `vendor\clink.lua` checks the Git config for  `cmder.status` and `cmder.cmdstatus`  as a way to disable Git status checks for this very complaint.  You can look in the [README.md](https://github.com/cmderdev/cmder/blob/44f5095047c2e5056884eca90082e13ebcf73061/README.md#git-status-opt-out) for how to disable Git status checks both globally and per Git Repo.

chrisant996:
Yes.  But Cmder also includes a whole separate copy of the source control prompt filters, inside `vendor\clink-completions`.

The git prompt filter in `vendor\clink.lua` checks git config.
The git prompt filter in `vendor\clink-completions\git_prompt.lua` **_does not_** check git config.

itsagam:
> Yes. But Cmder also includes a whole separate copy of the source control prompt filters, inside `vendor\clink-completions`.
> 
> The git prompt filter in `vendor\clink.lua` checks git config.
> The git prompt filter in `vendor\clink-completions\git_prompt.lua` **_does not_** check git config.
> 
> So, the instructions and mechanism for opting out of git status aren't effective.

Can confirm. Just tried Cmder today and had to spent quite a while going down this rabbit hole. Disabled git status with the instructions from the README and tried multiple solutions from this thread https://github.com/cmderdev/cmder/issues/447#. They do disable the prompt but the delay is there. Clink.lua was iterating through all the lua files in the completions directory and executing them, one of which was another copy of the git_prompt_filter.

daxgames:
[...]

PR #2520, opened 2021-04-08
set global cmderGitStatusOptIn var for use in clink-completions

daxgames:
Requires the following two lines to be added to `%cmder_root%`vendor\clink-completions\git_prompt.lua`  function `load_ini(fileName)`:

```
local function load_ini(fileName)
    -- Check for Cmder configured Git Status Opt In/Out - See: https://github.com/cmderdev/cmder/issues/2484
    if cmderGitStatusOptIn == false then return nil end
```

This will allow Cmder Git Status Opt Out feature to work 100% speeding up return to prompt in large git repos where desired.

I will also be opening a PR in Clink Completions to get this change added there.

Resolves: #2484

### DIFF
--- a/vendor/clink.lua
+++ b/vendor/clink.lua
@@ -285,15 +285,24 @@ end
 -- @return {bool}
 ---
 local function get_git_status_setting()
-    gitStatusSetting = io.popen("git --no-pager config -l 2>nul")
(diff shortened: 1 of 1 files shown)


daxgames:
https://github.com/vladimir-kotikov/clink-completions/pull/141

PR #2537, opened 2021-05-02
Fix env symbol being repeated when using a program that sets/modifies the env

Lunchb0ne:
Refactor and separate the env substitution part of the Prompt.

image

This helps fix this weirdness that happens whenever the prompt is printed.

In the original implementation, the value `prompt_lambSymbol` has the value `"("..env..") "` appended to it all the time, but actually the variable is never reset, so it just keeps adding the env's name every time the prompt is printed.

### DIFF
--- a/vendor/clink.lua
+++ b/vendor/clink.lua
@@ -83,14 +83,13 @@ local function set_prompt_filter()
       cr = ' '
     end
 
-    if env ~= nil then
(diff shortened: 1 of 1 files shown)


daxgames:
@Lunchb0ne  Merged - Thanks for the PR

PR #2549, opened 2021-05-21
Show branch name when cmder.status=false without expensive status ops

### LINKED ISSUE #2548: Git status slows down prompt in large repos
### Purpose of the issue
- [ ] Bug report (encountered problems/errors)
- [x] Feature request (request for new functionality)
- [ ] Question

### Version Information

191012 [x64] {Preview}

Windows 10 Enterprise 21H1

### Description of the issue

Running `git status` and `git diff` in large repos can take over 1-2 seconds each. By default, cmder runs both in series when generating the prompt text after every command.

So, running fast commands like `cd ..` in large repos can easily take 3-4 seconds for the prompt to come back.

Cmder already has a feature to disable the git status in the prompt to fix this by adding the following to your `.git/config`
```
[cmder]
    status = false
```

This is obviously faster, but I would really like to see the branch name still, and that doesn't require running either `git status` or `git diff` so doesn't slow down in large repos.

### Proposal
Support an additional config option
```
[cmder]
    status = branchonly
```

Update `get_git_status_setting` to return `true | false | 'branchonly'`, and in `git_prompt_filter` check for `'branchonly'` before running `get_git_status` and `get_git_conflict` to decide on the color.

This change is backward compatible, and pretty small. I am happy to submit a PR to contribute this.

ian-craig:
Fixes #2548

Adds support for skipping status check (git status and git diff) when generating the prompt, while still showing the branch name.

Adds a simple mode for powershell and bash which get the branch name by just reading `.git/HEAD` which should not slow down with large repos.

I decided to change the color to gray when `status=false` to indicate that the status is unknown. This would avoid misleading people into thinking they are "clean" if they apply the config only to specific repos.

### DIFF
--- a/vendor/clink.lua
+++ b/vendor/clink.lua
@@ -365,29 +365,27 @@ end
 
 ---
 -- Get the status of working dir
--- @return {bool|'branchonly'}
+-- @return {bool}
 ---
 local function get_git_status_setting()
-    local gitCmdStatusConfig = io.popen("git --no-pager config cmder.cmdstatus 2>nul")
-    for line in gitCmdStatusConfig:lines() do
-        if string.match(line, 'false') then
-          gitCmdStatusConfig:close()
-          return false
-        elseif string.match(line, 'branchonly') then
-          gitCmdStatusConfig:close()
-          return 'branchonly'
-        end
-    end
-    gitCmdStatusConfig:close()
-
     local gitStatusConfig = io.popen("git --no-pager config cmder.status 2>nul")
+
     for line in gitStatusConfig:lines() do
         if string.match(line, 'false') then
           gitStatusConfig:close()
           return false
         end
     end
+
+    local gitCmdStatusConfig = io.popen("git --no-pager config cmder.cmdstatus 2>nul")
+    for line in gitCmdStatusConfig:lines() do
+        if string.match(line, 'false') then
+          gitCmdStatusConfig:close()
+          return false
+        end
+    end
     gitStatusConfig:close()
+    gitCmdStatusConfig:close()
 
     return true
 end
@@ -404,13 +402,13 @@ local function git_prompt_filter()
 
     local git_dir = get_git_dir()
     cmderGitStatusOptIn = get_git_status_setting()
-    if cmderGitStatusOptIn then
-      if git_dir then
-          -- if we're inside of git repo then try to detect current branch
-          local branch = get_git_branch(git_dir)
-          local color
-          if branch then
-              if cmderGitStatusOptIn ~= 'branchonly' then
+
+    if git_dir then
+        -- if we're inside of git repo then try to detect current branch
+        local branch = get_git_branch(git_dir)
+        local color = colors.unknown
+        if branch then
+            if cmderGitStatusOptIn then
                 -- Has branch => therefore it is a git folder, now figure out status
                 local gitStatus = get_git_status()
                 local gitConflict = get_git_conflict()
@@ -423,14 +421,11 @@ local function git_prompt_filter()
                 if gitConflict then
                     color = colors.conflict
                 end
-              else
-                color = colors.unknown
-              end
+            end
 
-              clink.prompt.value = string.gsub(clink.prompt.value, "{git}", color.."("..verbatim(branch)..")")
-              return false
-          end
-      end
+            clink.prompt.value = string.gsub(clink.prompt.value, "{git}", color.."("..verbatim(branch)..")")
+            return false
+        end
     end
 
     -- No git present or not in git file
--- a/vendor/clink.lua
+++ b/vendor/clink.lua
@@ -41,6 +41,10 @@ local function get_conflict_color()
   return conflict_color or "\x1b[31;1m"
 end
 
+local function get_unknown_color()
+  return unknown_color or "\x1b[30;1m"
+end
+
 ---
 -- Makes a string safe to use as the replacement in string.gsub
 ---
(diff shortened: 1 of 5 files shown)


vendor/clink.lua
```diff
@@ -361,27 +365,29 @@ end
 
 ---
 -- Get the status of working dir
--- @return {bool}
+-- @return {bool|'branchonly'}
```
chrisant996, line 368: Making the return type polymorphic adds extra burden on call sites, and adds potential for regression in unexpected situations.

vendor/clink.lua
```diff
@@ -403,17 +410,21 @@ local function git_prompt_filter()
           local branch = get_git_branch(git_dir)
           local color
           if branch then
-              -- Has branch => therefore it is a git folder, now figure out status
-              local gitStatus = get_git_status()
-              local gitConflict = get_git_conflict()
+              if cmderGitStatusOptIn ~= 'branchonly' then
```
chrisant996, line 413: The [vladimir-kotikov/clink-completions](https://github.com/vladimir-kotikov/clink-completions) repo has a dependency on `cmderGitStatusOptIn` being a true/false value.  The contents of that repo are copied into the %CMDER_ROOT%\vendor\clink-completions directory.

This new setting won't behave as intended in certain repos, and when it's used it will bring back the slow behavior https://github.com/cmderdev/cmder/issues/2484 that was fixed by https://github.com/cmderdev/cmder/commit/6027ac31c428a3b9a5cf12424d4d95787e279023.

chrisant996 (changes requested):
Overall it looks cool.  If I'm reading it right, then the way that `cmderGitStatusOptIn` now can be a bool or a string seems problematic (see specific comments for details).

A couple alternatives could be:
- Consider updating the usage of `cmderGitStatusOptIn` in the clink-completions repo as well.
- Consider making a new config setting instead of modifying an existing one.

Or @daxgames may have something else in mind.

vendor/clink.lua
ian-craig: Thanks, wasn't aware of this.

Would the following seem OK instead?

```
local statusSetting = get_git_status_setting()
cmderGitStatusOptIn = statusSetting == true
...
if statusSetting ~= 'branchonly' then
```

I'll be unavailable for the next few days but can come back and iterate on this next week.

ian-craig: Although.. clink-completions just does `cmderGitStatusOptIn == false` so `'branchonly' == false` would be false, which may actually be what we want here anyway. I could still do `cmderGitStatusOptIn = statusSetting ~= false` to make this more explicit on our end.

But my question on that is... do we want clink-completions's filter to still run. The goal of this PR was to avoid expensive operations, and the clink-completions filter appears to only be doing fast operations that aren't affected by repo size.

daxgames: I am not aware of anyone ever complaining we showed the Git branch in the prompt.  The only complaint has ALWAYS been the prompt can be slow in a large Git repo.  

My thought was that `*status = false` changes from 'no status and no branch' to 'no status with branch' so we scrap the `cmdstatus = branchonly` setting.

Currently Cmder's Git Opt Out functionality has feature parity across `Cmd`, `Powershell`, and `Bash` shells this PR only affects git status opt out in `Cmd` shells so we should also:

- Modify `%cmder_root%`\vendor\git-prompt.sh  to just display the branch for Bash shells if `status=false` or `bashstatus=false`
- Modify `%cmder_root%`\vendor\psmodules\Cmder.ps1`  to just display the branch for Powershell shells if `status=false` or `psstatus=false`
  -  We do not have much control of how Powershell prompt is displayed because we use a function from `Posh-Git` so this  will be more difficult.

This would maintain current compatibility with `bash` and `powershell` shells for these settings. 

Lastly disabling git status checks in Cmder whether the branch is displayed or not the clink-completions git-prompt filter should not run either.

chrisant996: > the clink-completions filter appears to only be doing fast operations that aren't affected by repo size.

It only seems that way because most of us don't use repos large enough to reproduce the problem. I recommend reading #2484, and especially [this comment](https://github.com/cmderdev/cmder/issues/2484#issuecomment-787043078).

ian-craig: This all sounds good to me, so I've:
1. Removed the "branchonly" concept
2. Made `cmder.*status=false` only prevent the status operations, not printing branch name for cmd
3. Added new simple branch name logic for bash and powershell to mimic the behavior of cmd in the `status=false` case

ian-craig: (And I'm no longer touching anything related to clink-completions so that is still off when `status=false`)

chrisant996: The .lua part looks good to me now.  I have no familiarity with the affected .sh or .ps1 scripts or how Cmder integrates with those shells, though.

daxgames: Will test it out later.  Thanks

daxgames:
Does not seem to work for me:

`cmd::cmder`
image

`bash::bash`
image

`powershell::powershell`
image

Powershell seems to work but neither `cmd::cmder` not `bash::bash`  display the branch.

daxgames:
@ian-craig I opened a PR to your branch that fixes the `cmd::Cmder` side of things:

https://github.com/ian-craig/cmder/pull/1

image

ian-craig:
Thanks @daxgames, I've merged your changes and am happy with the switch to white.

Unrelated side note... Is there supposed to be no space before the branch name in cmd? It looks like there used to be a space in older versions for cmd, and still is for powershell and bash. I thought I'd messed something up in my testing but it looks like your screenshots show this too.

PR #2556, opened 2021-06-06
Clink async prompt

chrisant996:
# Description
This PR adds support in Cmder for the new "asynchronous prompt filtering" in Clink v1.2.10 and higher.  It should be forward/backward compatible with other versions of Clink.

Now the potentially slow `git status` (etc) commands are run in the background, so there's no waiting for the prompt to show up.  When the background commands finish, the prompt gets refreshed.  To avoid unnecessary flicker, the most recent results are cached so they can be used while waiting for new background commands to finish (switching repos or branches resets the cached results).

The new async prompt filtering is enabled by default in Clink, but prompt filters must opt in to use it (which is what this PR does for Cmder's git prompt).  It can be turned off via `clink set prompt.async false`.  See [Asynchronous Prompt Filtering](https://chrisant996.github.io/clink/clink.html#asyncpromptfiltering) for more info.

# Merge Notes
This PR is based on the pending PR https://github.com/cmderdev/cmder/pull/2549.  When applied _after_ https://github.com/cmderdev/cmder/pull/2549 is committed, I expect this PR should encounter few-or-no merge conflicts.  But if it does encounter merge conflicts, let me know and I can resolve them.

# Test Notes
- [x] Tested Cmder 1.3.18 with this change and Clink v1.1.45 (what it currently includes -- which is still a pre-release build of Clink).
- [x] Tested Cmder 1.3.18 with this change and Clink v1.2.10 (latest release build).
  - [x] Tested with `prompt.async` setting **true** (the default).
  - [x] Tested with `prompt.async` setting **false**.

Everything looks good to me.

### DIFF
--- a/vendor/clink.lua
+++ b/vendor/clink.lua
@@ -86,6 +86,13 @@ end
 
 
 ---
+-- Global variable so other Lua scripts can detect whether they're in a Cmder
+-- shell session.
+---
+CMDER_SESSION = true
+
+
+---
 -- Setting the prompt in clink means that commands which rewrite the prompt do
 -- not destroy our own prompt. It also means that started cmds (or batch files
 -- which echo) don't get the ugly '{lamb}' shown.
@@ -135,6 +142,10 @@ local function set_prompt_filter()
       prompt_singleLine = false
     end
 
+    if prompt_includeVersionControl == nil then
+      prompt_includeVersionControl = true
+    end
+
     if prompt_type == 'folder' then
         cwd = get_folder_name(cwd)
     end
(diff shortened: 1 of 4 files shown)


vendor/clink.lua
```diff
@@ -59,6 +63,28 @@ local function get_folder_name(path)
...
+---
+local io_popenyield
+local clink_promptcoroutine
+local cached_info = {}
+if clink.promptcoroutine and io.popenyield then
+    io_popenyield = io.popenyield
+    clink_promptcoroutine = clink.promptcoroutine
+    -- Uncommenting this will override the cmderGitStatusOptIn and always show
+    -- git status when Clink is able to run it in the background.
+    --cmderForceAsyncGitStatus = true
```
daxgames, line 79: could this be a config setting in the https://github.com/cmderdev/cmder/blob/master/vendor/cmder_prompt_config.lua.default

chrisant996: Sure, I'll add it!  I wasn't sure whether/where/how to make it configurable.  :)

chrisant996: Done.  Do the name and description seem reasonable?

daxgames: @chrisant996 Why set a setting that just controls setting another setting?  Why not just set the setting?

chrisant996: The intent was to decouple the user-configurable setting from an internal programmatic variable that the Cmder code may set to override the user-configurable setting if necessary.

But I can remove that, no problem.

chrisant996: It is crucial for correct functionality:

- `prompt_overrideGitStatusOptIn` is a user-configurable setting.
- `cmderForceAsyncGitStatus` is a programmatically set variable, and it must not respect the user-configurable setting when using a version of Clink that does not support async prompt filtering.

I can make `cmderForceAsyncGitStatus` be a `local` variable, so that it doesn't feel like a setting.

daxgames: Cool it was just a question.  I was sure you had a reason to do it.  Glad its better now.

daxgames:
@stanzilla thanks I was just starting to look at this yesterday.

daxgames:
I am sorry this took me so long to look at this PR.  This is an awesome feature, speeds up the prompt to the point the delay is all but imperceptible!  Thank you @chrisant996 for this and all the work you have done on revitalizing `clink`

PR #2569, opened 2021-07-12
Update clink to 1.2.22

AltarBeastiful:
Latest update of clink contains some interesting changes. 
Mainly the ability to [complete on doskey aliases ](https://github.com/chrisant996/clink/issues/119)

### DIFF
--- a/vendor/sources.json
+++ b/vendor/sources.json
@@ -6,8 +6,8 @@
     },
     {
         "name": "clink",
-        "version": "1.2.20",
-        "url": "https://github.com/chrisant996/clink/releases/download/v1.2.20/clink.1.2.20.801802.zip"
+        "version": "1.2.22",
+        "url": "https://github.com/chrisant996/clink/releases/download/v1.2.22/clink.1.2.22.890a84.zip"
     },
     {
         "name": "conemu-maximus5",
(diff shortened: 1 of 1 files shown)


chrisant996:
I recommend updating to Clink v1.2.22 when it gets released in the next day or two.
Some highlights:
- Git prompt does its expensive work in the background, nearly eliminating the "prompt lag".
- New `clink-select-complete` command that has Powershell-like and Fish-like interactive completion.
- Shift+arrow keys select input text, typing replaces selection, etc.
- Argmatchers for commands work even in doskey aliases that use the commands.
- Many bug fixes since the v1.1.45 Cmder currently includes.
- v1.2.22 will complete a major milestone, and finish the planned feature work for a while.

chrisant996:
Clink v1.2.22 has been released.  I recommend updating Cmder to include that instead of 1.2.20, because it includes a few additional fixes and features.

AltarBeastiful:
Updated PR with latest version as proposed

chrisant996:
Clink v1.2.23 includes two minor fixes, sorry.  Not a big deal if it's not picked up in the next Cmder yet, but <kbd>PgDn</kbd> could behave strangely in `clink-select-complete`, and a purely cosmetic change for the key name for <kbd>Ctrl</kbd>+<kbd>M</kbd> when `terminal.differentiate_keys` was false.

PR #2609, opened 2021-09-25
Fix no env in the prompt string

Lunchb0ne:
Yeah, I was surprised too..
The `{env}` was missing in the prompt constructor, it wasn't there even after subbing.

(Rebased it again, sorry I didn't see it before)

### DIFF
--- a/vendor/clink.lua
+++ b/vendor/clink.lua
@@ -168,7 +168,7 @@ local function set_prompt_filter()
 
     local version_control = prompt_includeVersionControl and "{git}{hg}{svn}" or ""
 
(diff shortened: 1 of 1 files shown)


daxgames:
@Lunchb0ne I don't use a lot of Python on Windows and thus do not use virtual environments either.  Could you provide a screenshot of this commit working so I can see what it is supposed to look like.

Lunchb0ne:
image

This is how the env is supposed to be displayed whenever you activate it.

daxgames:
Merged - Thanks

PR #2618, opened 2021-10-29
Fix a problem with an ampersand character in path

xpam-net:
If the path variable contained values with an ampersand character (such as in the case of MySQL), the string splits by this character, and tries to execute what follows as a separate command. 
All occurrences of the set command containing %PATH% should be wrapped in quotation marks

### DIFF
--- a/vendor/lib/lib_path.cmd
+++ b/vendor/lib/lib_path.cmd
@@ -74,7 +74,7 @@ exit /b
     set "find_query=%add_to_path%"
     set "find_query=%find_query:\=\\%"
     set "find_query=%find_query: =\ %"
-    set OLD_PATH=%PATH%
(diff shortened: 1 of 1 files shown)


vendor/lib/lib_path.cmd
```diff
@@ -121,8 +121,8 @@ exit /b
     exit /b
 
     :toolong
-      echo %OLD_PATH%>tempfileA
-      echo %PATH%>tempfileB
+      echo "%OLD_PATH%">tempfileA
+      echo "%PATH%">tempfileB
```
asoluter, line 125: Echo outputs quotes to file as well here

daxgames:
I merged then reverted this based on the review comment above.

If this still needs to be done fix if necessary and re-open a new PR.

PR #2621, opened 2021-11-08
Speed up git status / conflict status by calling git only once.

chehrlic:
Avoid to call git two times when trying to get the (conflict)status by examining 'git status' flags as described here: https://git-scm.com/docs/git-status
Currently there is a small behavior change - unversioned files are ignored instead leading to a dirty status.

### DIFF
--- a/vendor/clink.lua
+++ b/vendor/clink.lua
@@ -339,32 +339,27 @@ local function get_svn_branch(svn_dir)
 end
 
 ---
--- Get the status of working dir
--- @return {bool}
+-- Get the status and conflict status of working dir
+-- @return {bool <status>, bool <is_conflict>}
 ---
 local function get_git_status()
     local file = io_popenyield("git --no-optional-locks status --porcelain 2>nul")
+    local conflict_found = false
+    local is_status = true
     for line in file:lines() do
-        file:close()
-        return false
+        local code = line:sub(1, 2)
+        -- print (string.format("code: %s, line: %s", code, line))
+        if code == "DD" or code == "AU" or code == "UD" or code == "UA" or code == "DU" or code == "AA" or code == "UU" then
+          is_status = false
+          conflict_found = true
+          break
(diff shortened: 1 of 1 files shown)


vendor/clink.lua
```diff
@@ -339,32 +339,27 @@ local function get_svn_branch(svn_dir)
...
--- @return {bool} indicating true for conflict, false for no conflicts
----
-function get_git_conflict()
-    local file = io_popenyield("git diff --name-only --diff-filter=U 2>nul")
-    for line in file:lines() do
-        file:close()
-        return true;
+        local code = line:sub(1, 2)
+        -- print (string.format("code: %s, line: %s", code, line))
+        if code == "DD" or code == "AU" or code == "UD" or code == "UA" or code == "DU" or code == "AA" or code == "UU" then
```
daxgames, line 352: Looks like you may have missed a couple of combinations see below from [here](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---name-only):

image

Unless I am not understanding this documentation correctly

chehrlic: But I only look for conflicted files/directories here - therefore unmerged should be enough.

chehrlic: Sorry, misread your message. I'm not using git --diff but git status: https://git-scm.com/docs/git-status#_short_format

daxgames: I understand that but my response is based on what the original `git diff --diff-filter=U` code did.  In my very unclear way, I was questioning how ` git diff --diff-filter=U`  would produse a status marker of `DD`.

I have since found other sources that claim `--diff-filter=U` produces statuses of `DD, AU, UD, UA, DU, AA, UU`

daxgames:
Thanks,

Merged

vendor/clink.lua
chehrlic: Ah sorry - then I completely misunderstood you :)

PR #2688, opened 2022-04-16
Remove set term=cygwin in init.bat

daxgames:


daxgames:
I have tested this and it resolves some long standing issues that make vim in CMD.exe based Cmder sessions unuseable if not dangerous in that it introduced random characters if navigating using arrow keys in vim while editing code.  I do not know why `set term=cygwin` was in the code since it was here before I got involved.

Speaking of other things that predate me we really should talk about losing the  'λ' it is the root of MANY issues.  I have read past objections to this idea but I think it is time.

MartiUK (approved):

MartiUK:
Yeah looks like it was some vestige of when git-for-windows was first integrated. LGTM 👍🏻

Issue #2702, opened 2022-05-24
On large svn repo prompt with svn status enabled can be slow


Mikaz-fr:
Similarly to #2548 and #1566, on large svn repo fetching the status delay slightly the command prompt display.

I recently updated to Cmder 1.3.19 and was thrilled to see that git status can now be fetched asynchronously using recent clink feature (integrated via #2556). 
Is there any plan to provide the same for svn ?

It's good now we can disable in the option the svn status, but having it asynchronously would be even better.

chrisant996:
There wasn't any plan to do so, because no one had identified it as being a problem that needed solving.  🙃

It's not hard to do.  Someone could make the Lua script tweaks and submit a PR.

(I'll see about extending [clink-flex-prompt](https://github.com/chrisant996/clink-flex-prompt) accordingly.  Someone could port from there to Cmder.  Or if I find myself with extra time, then maybe at some point I can also provide a patch for Cmder, but that's not a priority for me.)

Mikaz-fr:
I've made an attempt to solve this for svn similarly to how it's done for git. I've open pull request #2703.
Please let me know if you would have done that differently.

chrisant996:
The changes look good to me!

Mikaz-fr:
Awesome. Feel free to close once PR is merged. Will get the next release with this feature :)

Mikaz-fr:
This change is now part of release https://github.com/cmderdev/cmder/releases/tag/v1.3.20.
This issue can be closed.

PR #2703, opened 2022-05-25
Add optional clink async prompt update for svn status

Mikaz-fr:
As discussed in #2702, svn status can take quite some time. This change aims to optionally perform svn status check asynchronously, like it's provided for git status.

In the setting file I've move `prompt_includeVersionControl` up as it makes sense to decide on that before deciding if sync or async update should be used. I've added a new setting for svn status update.
In `clink.lua` this new setting is used to decide if the status should be checked asynchronously or blocking. Similar caching mechanism than git is used too.

I've done limited testing with the different svn repo I have and it seems to behave. Please propose improvement if needed.

### DIFF
--- a/vendor/clink.lua
+++ b/vendor/clink.lua
@@ -338,7 +338,7 @@ end
 -- @return {false|svn branch name}
 ---
 local function get_svn_branch(svn_dir)
-    local file = io.popen("svn info 2>nul")
+    local file = io_popenyield("svn info 2>nul")
     for line in file:lines() do
         local m = line:match("^Relative URL:")
         if m then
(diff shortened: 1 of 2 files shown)


chrisant996 (commented):
There is only [one prompt coroutine](https://chrisant996.github.io/clink/clink.html#clink.promptcoroutine) per prompt filter.  One coroutine is shared by `git` and `svn` (and `hg` etc), so if both git and svn call `clink_promptcoroutine` then whichever happens first specifies the function that will actually run.  So, if a directory is part of both git and svn (unlikely, but possible) then the prompt will malfunction.

That's an unlikely scenario for git vs svn, maybe, but in the future other prompt segments may want to run async as well, and they may need to coexist with git and/or svn.

So, the `clink_promptcoroutine` stuff needs to be rearranged a bit.  I can do that tonight.

(I already updated [clink-flex-prompt](https://github.com/chrisant996/clink-flex-prompt).)

PR #2709, opened 2022-06-09
Update Clink to 1.3.19

bachig26:


bachig26:
i'm closing this, as i think the frequent maintainers will do them, as i think it is not going to anytime soon.

Issue #2757, opened 2022-10-16
`./build.ps1 -verbose` fails looking for `msbuild.exe`


daxgames:
### Purpose of the issue
- [X] Bug report (encountered problems/errors)
- [ ] Feature request (request for new functionality)
- [ ] Question

### Version Information

Cmder: master

### Description of the issue

`./build.ps1 -verbose` fails looking for `msbuild.exe` if running from a simple Powershell session that is not `msbuild` aware.  This should only happen if the `-compile` switch is also used.

DRSDavidSoft:
@daxgames Thanks, PR #2758 will fix this issue.

daxgames:
@DRSDavidSoft You are too fast.  🥇  Initially just documenting so I would not forget it but I was fixing after seeing the issue in the code.

PR #2758, opened 2022-10-16
Re-factor the build script to distinguish each step

DRSDavidSoft:
This PR addresses #2757

### Description
- Added a `-noVendor` switch to bypass downloading vendor packages
- Only check for `msbuild.exe` in path when the `-Compile` switch is used

### DIFF
--- a/.github/workflows/vendor.yml
+++ b/.github/workflows/vendor.yml
@@ -1,0 +1,40 @@
+name: Update Vendor
+
+on:
+  workflow_dispatch:
+  schedule:
+  # At 13:37 UTC every day.
+  - cron: '37 13 * * *'
+  
+defaults:
+  run:
+    shell: pwsh
+    
+permissions:
+  contents: read
+
+jobs:
+  vendor:
+
+    runs-on: windows-latest
+    continue-on-error: false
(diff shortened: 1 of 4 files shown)


DRSDavidSoft:
@daxgames Please have a look, thanks!

daxgames:
@DRSDavidSoft are you adding Github `gh` as a Cmder component?  If so I like it.

DRSDavidSoft:
> @DRSDavidSoft are you adding Github `gh` as a Cmder component? If so I like it.

That's also a good idea, but those functions are for the GitHub Actions / CI.  I think `gh` could also be useful since it's only ~8 MB, but not the focus of this PR.

daxgames:
I could see adding gh to the `%CMDER_ROOT%/vendor/bin folder so it is automatically in the path.  At this point it is easy enough to add after by a user that wants it.

daxgames:
Odd it downloads from Github actions as `Cmder.zip.zip`

DRSDavidSoft:
@daxgames Sad to say, this is a known limitation by the authors (https://github.com/actions/upload-artifact#zipped-artifact-downloads).  I haven't found anything yet to overcome this

PR #2759, opened 2022-10-16
do not fail on msbuild if not compiling

### LINKED ISSUE #2757: `./build.ps1 -verbose` fails looking for `msbuild.exe`

daxgames:
Fix #2757


DRSDavidSoft:
Thanks for closing as the Scoop installer was inadvertently added.  Also #2758 is more robust and has improved readability.

daxgames:
@DRSDavidSoft I am not sure where the Scoop installer came from.

PR #2765, opened 2022-10-18
Git-bash use conemu-msys2-64 connector

nexplorer-3e:
use conemu-msys2-64 connector would help to solve some display problems under Git Bash like 
 - character drifting, 
 - cursor wrong position,
 - `tmux` display wrong, 
 - and so on. 
As the default profile is hardcoded and ConEmu upstream has solved it by using msys-connector, I copied from ConEmu default profile to use it.

### DIFF
--- a/vendor/ConEmu.xml.default
+++ b/vendor/ConEmu.xml.default
@@ -549,14 +549,14 @@
 					<value name="GuiArgs" type="string" data=" /icon &quot;%CMDER_ROOT%\icons\cmder.ico&quot;"/>
 					<value name="Active" type="long" data="0"/>
 					<value name="Count" type="long" data="1"/>
-					<value name="Cmd1" type="string" data="*&quot;%ConEmuDir%\..\git-for-windows\usr\bin\bash.exe&quot; --login -i"/>
+					<value name="Cmd1" type="string" data="*set &quot;PATH=%ConEmuDir%\..\git-for-windows\usr\bin;%PATH%&quot; &amp; %ConEmuDir%\..\git-for-windows\git-cmd.exe --no-cd --command=%ConEmuBaseDirShort%\conemu-msys2-64.exe &quot;%ConEmuDir%\..\git-for-windows\usr\bin\bash.exe&quot; --login -i"/>
 				</key>
 				<key name="Task8" modified="2018-02-22 06:05:13" build="171109">
 					<value name="Name" type="string" data="{bash::bash}"/>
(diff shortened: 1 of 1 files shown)


nexplorer-3e:
Another thing about using conemu connector is `xterm-256color`. For example, as shown in [ConEmu Docs](https://conemu.github.io/en/AnsiEscapeCodes.html#Example_1_Vim), show real color of `vim` need to add `-cur_console:h0` startup parameter to make it work. But turn BufferHeight off has other sides, for which I didn't add it.

DRSDavidSoft:
@nexplorer-3e I merged the changes for additional testing in Cmder.  Thanks for your contribution!

@daxgames Can you please test it out?

daxgames:
@DRSDavidSoft I generally try and test these out before merging to master as I assume you did, and you just wanted a second set of eyes on it.  Seems to work fine.

I never experienced any of the issues it is supposed to fix so I see no difference in how it works so not completely clear on the benefits other than @Maximus5 always brings up the connector when any display issues are mentioned in ConEmu issues.  This will help with Cmder bash and those situations. 

One note this change only affects new installs of Cmder and existing users will need to update their bash tasks as follows in the Conemu GUI:

1. `bash::bash`
   a. Update `"%ConEmuDir%\..\git-for-windows\usr\bin\bash.exe"; --login -i`
   b. To `set "PATH=%ConEmuDir%\..\git-for-windows\usr\bin;%PATH%" & %ConEmuDir%\..\git-for-windows\git-cmd.exe --no-cd --command=%ConEmuBaseDirShort%\conemu-msys2-64.exe "%ConEmuDir%\..\git-for-windows\usr\bin\bash.exe" --login -i`
1. `bash::bash as Admin`
   a. Update `*"%ConEmuDir%\..\git-for-windows\usr\bin\bash.exe"; --login -i`
   b. To `*set "PATH=%ConEmuDir%\..\git-for-windows\usr\bin;%PATH%" & %ConEmuDir%\..\git-for-windows\git-cmd.exe --no-cd --command=%ConEmuBaseDirShort%\conemu-msys2-64.exe "%ConEmuDir%\..\git-for-windows\usr\bin\bash.exe" --login -i`

DRSDavidSoft:
Thanks, as you said I tested (only for potential issues) and didn't notice any problems, however as you said I didn't notice these problems either.  Since this change was introduced in upstream, I believe using ConEmu connectors hopefully won't be an issue.

PR #2790, opened 2022-12-03
hide update checks clink in CMD integration

candrapersada:


DRSDavidSoft:
We should also turn off ConEmu updates, if clink's is off.

I'd feel comfortable merging this after we provide an auto update mechanism for Cmder and it's vendored dependencies.

candrapersada:
what I mean is hide update checks clink in CMD
image

daxgames:
I think this should be configurable somehow.  I actually like the clink auto update check feature and like the default.

Not going to merge this just yet.

daxgames:
I think the default should be true here. and if the user wants to change it they can set it to false.

DRSDavidSoft:
I'm going to close this, because:

1. It's a good idea to regularly update Clink when there are newer versions.
2. If a user doesn't want this, Clink has clear instructions on how to disable update checks right in the message.
3. If we are going to disable auto-updates, I'd rather the settings be consistent with ConEmu also.

For reference, I personally _do_ disable auto clink updates and ConEmu updates as well, because I don't like the upstream packages to be updated in front of Cmder, and I find the update messages annoying as well. But I'd prefer not to enforce this for all of the users.

Disabling auto updates globally might also pose some security risks. It's better if the user applies such configurations if they choose to do so.

Issue #2789, opened 2022-12-01
[Bug] 连续ctrl+c ，窗口卡死


siruoren:
### Version Information

```markdown
Cmder version: 1.3.20
Operating system: win0 专业版
```

### Cmder Edition

Cmder Full (with Git)

### Description of the issue

[Bug] 连续ctrl+c ，窗口卡死

### How to reproduce

连续ctrl+c 

### Checklist

- [X] I have read the documentation.
- [X] I have searched for similar issues and found none that describe my issue.
- [X] I have reproduced the issue on the latest version of Cmder.
- [X] I am certain my issues are not related to ConEmu, Clink, or other third-party tools that Cmder uses.

DRSDavidSoft:
对不起，不清楚你在说什么。Post an screenshot of the problem if possible.

siruoren:
> 对不起，不清楚你在说什么。Post an screenshot of the problem if possible.

![image](https://user-images.githubusercontent.com/28772488/205607459-c0df4144-1615-4a74-8b75-6b7c72f3ec24.png)
连续ctrl+c，然后就会卡在当前状态

siruoren:
> What is your Surface device? Arm or x64 Intel?

 x64 Intel

DRSDavidSoft:
Thanks, please write in English, and can you capture a video, I still can not understand what is happening.
You can write more descriptive messages in Chinese if you prefer.
thanks.

chrisant996:
@DRSDavidSoft what @siruoren is reporting is that holding down <kbd>Ctrl</kbd>-<kbd>C</kbd> eventually causes a hang in the cmd shell.

I am able to reproduce that inside Cmder with the 1.3.20 release (which is the current at the moment).

I attached a debugger to the cmd.exe process, and the hang is occurring because the <kbd>Ctrl</kbd>-<kbd>C</kbd> occurred while the prompt filter was invoking git, and the background git.exe process has become stuck within a `git.exe  --no-pager config cmder.cmdstatus` command.  Killing the hidden git.exe allows cmd.exe to regain control.

Some observations:
[...]

PR #2791, opened 2022-12-05
Fix https://github.com/cmderdev/cmder/issues/2789

chrisant996:
Holding ^C made git.exe hang while cmd.exe (Clink) updated the prompt.

The prompt script had three problems:

1.  It invoked `git config` every time a prompt was displayed, to figure out where to skip invoking `git status`.  But it even did that if the current directory wasn't part of a git repo.
2.  It invoked `git config` two times for every single prompt, to attempt to improve performance if the user disables `git status` coloring.  But two times for every single prompt is expensive, so it has the opposite effect in the general case, and noticeably degrades performance.
3.  It invoked `git config` using a blocking call, instead of using the async prompt support in Clink.  That significantly reduced the benefit of having used async prompt filtering for `git status`.

Now the `git config` invocations use async prompt filtering, which lets the prompt display instantaneously.  It also now uses a timer to avoid invoking `git config` repeatedly when new prompts show up in rapid succession.

Also, the `cmderGitStatusOptIn` variable is no longer leaked into the Lua global namespace.

These changes resolve the issue: holding ^C is very fast and no longer causes git.exe to hang.

### DIFF
--- a/vendor/clink.lua
+++ b/vendor/clink.lua
@@ -409,12 +409,62 @@ local function get_svn_status()
 end
 
 ---
+-- Get the status of working dir
+-- @return {bool}
+---
+local last_git_status_time = nil
+local last_git_status_setting = true
+local function get_git_status_setting()
+    local time = os.clock()
+    local last_time = last_git_status_time
+    last_git_status_time = time
+    if last_time and time >= 0 and time - last_time < 10 then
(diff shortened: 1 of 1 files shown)


DRSDavidSoft:
👍🏻 LGTM!

PR #2799, opened 2022-12-20
Fix error from profile.d PowerShell scripts

bdukes:
When there are `ps1` files in `profile.d`, the following error is logged:

>A positional parameter cannot be found that accepts argument 'Sourcing'.


DRSDavidSoft:
Thanks! 👍🏻

PR #2857, opened 2023-07-04
Integration scripts

daxgames:
@DRSDavidSoft New branch for us to work on integration scripts.  It already has all the integration scripts in `vendor/bin` from the `vagrant+packer` branch.


DRSDavidSoft:
Amazing stuff 👍🏻 Would you agree to moving some of the scripts to `vendor\scripts` instead of `vendor\bin`? Even `vendor\bin\scripts` would be nice!

daxgames:
The scripts are where they are so theybwill be in the path when running Cmder from the distribution package because they work for our processes as well as for end users wanting to add cmder to other terminals.

DRSDavidSoft:
Great, I propose the following changes. If you aprove, I'll implement them at a later time.

1. Move some of the scripts to `/vendor/scripts` and leave the others in `/vendor/bin`
2. Update the other branch to reference the new paths
3. Add `/vendor/scripts` to be in PATH the same as `/vendor/bin`

(For now, I propose that every script that starts with `add-` should be moved to `/vendor/scripts`).

Additionally, I'd like to add two new scripts:

1. Adding to/removing from the context menu (like what the launcher does, I like your PowerShell approach more than the .exe one in integrations. Can also give the ability to check for presense)
2. A script that would set the CMDER_ROOT environment variable
4. A script akin to [add-font.ps1](https://github.com/kenijo/shark/blob/master/bin/add-font.ps1) from Shark, a fork of Cmder that would allow the user to quickly `wget` some monospace fonts and install them

TBH, these scripts seem very useful to me! And I'm excited to getting them in the main branch. I hope you also agree with the additions that I'm going to work on.

Thanks, Dax!

DRSDavidSoft:
P.S. So sorry for being late on the reviewing of your Awesome PRs. I'm also very excited for your speed PR, I'd like to review it sooner so we can merge it for the next release. Much appreciated! 👍🏻

daxgames:
Therenis already a script that does #2.

daxgames:
Not sure the rationale for having two separate folders both added to the path when one folder added to the path does EXACTLY the same thing.

DRSDavidSoft:
Ah, thank you. I was thinking of keeping the `/bin` more clear of clutter, and to be clear I meant that `/scripts` should be added for PATH in the vagrant branch, as IMHO I'd prefer if the `/scripts` weren't in PATH for all users. I know adding more length to PATH is not the best idea, but I was thinking it'd be worth it to keep the `/bin` directory more clean. I hope you'd agree!

daxgames:
I'm not that opinionated about it, and they don't NEED to be in the path.  Just trying to make it easy for users and I HATE typing.

Issue #2859, opened 2023-07-06
[Bug]  %20 special character issue while navigating


vivekdhiman:
### Version Information

```markdown
Cmder version: Latest
Operating system:
```

### Cmder Edition

Cmder Full (with Git)

### Description of the issue

I just tried using thill tools. I have repository which contains a folder say My%20Home, when I navigate like : cd "My%20Home" Below error shows up.

prompt filter failed:
...\cmder\vendor\clink.lua:175: invalid capture index
stack traceback:
        [C]: in function 'gsub'
        ...cmder\vendor\clink.lua:175: in function '?'

### Checklist

- [X] I have read the documentation.
- [X] I have searched for similar issues and found none that describe my issue.
- [X] I have reproduced the issue on the latest version of Cmder.
- [X] I am certain my issues are not related to ConEmu, Clink, or other third-party tools that Cmder uses.

chrisant996:
Confirmed.  It's a bug in Cmder's custom prompt, and it exists in several places.

```lua
    prompt = string.gsub(prompt, "{cwd}", cwd)
```

The [`string.gsub()`](https://www.lua.org/manual/5.2/manual.html#pdf-string.gsub) function says that the third parameter is the replacement string, and that `%` is an escape character.

A few of the calls to `gsub()` explicitly use escaped `%%` in the third parameter; those ones are ok.

Cmder has a `verbatim()` function that handles escaping the `%`, but only 3 places use it (and 2 of those places use it technically incorrectly since they assume ANSI escape codes will never contain `%` characters, which isn't technically guaranteed).

Places that use `gsub()` need to ensure the third parameter is either properly escaped by calling `verbatim()`, or that escaping is already present (e.g. inside `verbatim()` itself), or that escaping is guaranteed to never be needed (e.g. when passing a hard-coded string that doesn't contain any `%`).

For example,
```diff
-    prompt = string.gsub(prompt, "{cwd}", cwd)
+    prompt = string.gsub(prompt, "{cwd}", verbatim(cwd))
```

And,
```diff
-            clink.prompt.value = string.gsub(clink.prompt.value, "{git}", " "..color.."("..verbatim(branch)..")")
+            clink.prompt.value = string.gsub(clink.prompt.value, "{git}", verbatim(" "..color.."("..branch..")"))
```

And,
```diff
-        cwd = string.gsub(cwd, clink.get_env("HOME"), prompt_homeSymbol)
[...]

PR #2860, opened 2023-07-09
clink.lua updates

daxgames:
# Changes from #2856 and #2859

## Adds 

- `%cmder_root%\vendor\clink.lua
  - Comments to explain some of less than clear purpose of some code.
  - Set global variable `CMDER_CONFIG_PATH=%CMDER_ROOT%\config`. 
  - Set global variable `CMDER_USER_CONFIG_PATH` to the value of env var '%CMDER_USER_CONFIG%' IF is is set else `nil`.
  - Wrap `gsub` args 2 and 3 with `verbatim()`


daxgames:
@chrisant996 I am not a lua dev so be gentle. :-)

I think I captured what you were talking about here.

vendor/clink.lua
```diff
@@ -666,10 +677,15 @@ for _,lua_module in ipairs(clink.find_files(completions_dir..'*.lua')) do
...
-    local cmder_config_dir = clink.get_env('CMDER_ROOT')..'/config/'
-    for _,lua_module in ipairs(clink.find_files(cmder_config_dir..'*.lua')) do
-        local filename = cmder_config_dir..lua_module
+---
+-- If Cmder is launched with '/c [folderPath]', indicating Cmder is installed globally and
+-- each user has a private '[folderPath]\config' folder, Clink won't know about the global
+-- 'CMDER_CONFIG_PATH` dir, so we need to load scripts from there before Clink loads lua 
+-- scripts from the profile directory given to it when it was injected.
+---
+if not (isempty(CMDER_USER_CONFIG_PATH)) then
```
chrisant996, line 686: I would recommend to add the comment, but don't change the code to use global Lua variables.  Using environment variables here is good.

I was recommending to add another environment variable in `init.bat` such as `set CMDER_DEFAULT_CONFIG_DIR=%CMDER_ROOT%\config`, and then use `%CMDER_DEFAULT_CONFIG_DIR%` in all the places instead of `%CMDER_ROOT%\config`.

daxgames:
WOW!  I thought I understood, did not know it was possible to be so wrong. :-)

chrisant996:
No worries! 😁

PR #2876, opened 2023-09-24
Ms2 winterm

daxgames:
# Windows Terminal Support

First looks at Windows Terminal Integration

image

So far I have:

- Added an argument to build.ps1 that allows building with Windows terminal instead of Conemu.
- Created a '%cmder_root%\vendor\WindowsTerminal_settings_default.json' that results in the WindowsTerminal Session and profiles shown above.
  - All sessions work as they do today in Cmder using Conemu.
  - Also includes changes to the word separator settings so it works more m=like we would expect it ti work.
  - NOTE: Non-Cmder sessions are auto-generated by Windows Terminal.
- Modified Cmder.bat to launch windows terminal by default if it exists and put the default config file where it needs to be before launching Windows Terminal


daxgames:
# Launcher works now.

Runs `%cmder_root%\vendor\windows-terminal\WindowsTerminal.exe` instead of ConEmu if `%cmder_root%\vendor\windows-terminal` exists.

- Supports all existing command line arguments but `/title`.
- Pass ALL arguments following `--` argument directly to `WindowsTerminal.exe` to use the full power of Windows Terminal from the command line.

PR #2898, opened 2023-11-21
Merging Cmder's `development` branch to `master` branch

DRSDavidSoft:
## Summary

This PR merges the current `development` branch on top of `master` and brings in the accumulated Cmder work since the last sync.

The biggest themes in this branch are:

- Windows Terminal support and related launcher/runtime updates
- A refactor of the build/pack scripts around package profiles
- Workflow/CI cleanup and consistency fixes
- A batch of shell, launcher, and configuration improvements
- Documentation and repo hygiene updates

### Windows Terminal support

This branch adds and refines support for launching Cmder in Windows Terminal, including:

- new/default Windows Terminal settings
- launcher updates in `Cmder.bat` and `launcher/src/CmderLauncher.cpp`
- new helper scripts for Git Bash, Mintty, and profile generation
- updates to shell integration files like `vendor/cmder.sh`, `vendor/profile.ps1`, and `vendor/psmodules/Cmder.ps1`
- support files for Windows Terminal defaults in `vendor/windows_terminal_default_settings.json`

### Build and packaging refactor

The build system has been modernized to use package profiles and shared helpers:

- new `scripts/package-profiles.json`
- new helper logic in `scripts/utils.ps1`
- updated `scripts/build.ps1`, `scripts/pack.ps1`, and `scripts/update.ps1`
- refreshed examples and documentation in `scripts/README.md`

This makes the packaging flow more structured and easier to extend.

### Workflow and CI updates

The GitHub Actions workflows were cleaned up and aligned with the new scripts:

- build workflow updates
- tests workflow updates
- CodeQL workflow updates
- vendor workflow updates
- summary formatting and artifact handling improvements

### Cmder launcher and runtime improvements

This branch also includes a substantial update to the launcher and core runtime behavior:

- launcher project and resource updates
- `CmderLauncher.cpp` changes for the new launch flow
- improvements to `Cmder.bat`
- updates to `vendor/init.bat`, `vendor/lib/lib_base.cmd`, and `vendor/lib/lib_path.cmd`
- shell startup and environment setup cleanup

### Documentation and repo hygiene

A number of documentation and maintenance updates are included as well:

- `README.md` and `config/Readme.md` adjustments
- `CHANGELOG.md` updates
- `.gitignore` and `packignore` additions
- `.git-blame-ignore-revs` for cleaner blame history

## Notes

This branch is mostly a consolidation of the Windows Terminal work, launcher/runtime improvements, and packaging refactors that have been developed on `development`.

## Verification

The updated workflows were run successfully on `development`, including:

- Build
- Tests
- CodeQL

## Files touched

This merge includes changes across:

- GitHub Actions workflows
- build and pack scripts
- launcher source
- Cmder startup and shell integration scripts
- Windows Terminal defaults and profile generation
- documentation and changelog files


DRSDavidSoft:
@copilot Hmm, I appreciate the detailed review, I'll address the issues raised in a local branch.

.github/workflows/build.yml
```diff
@@ -144,116 +136,67 @@ jobs:
...
-        path: build/hashes.txt
-        name: hashes.txt
-        archive: false
-
-    - name: Summary - Artifacts uploaded
+    - name: Summary - Package artifacts
       if: success()
       shell: pwsh
-      env:
-        GH_TOKEN: ${{ github.token }}
```
DRSDavidSoft, line 201: This block was removed because the workflow no longer uses `GH_TOKEN` for the summary step. The old code fetched artifact download URLs through the GitHub API, but this version just lists files directly from `build/`, so the token is unused here.

Since the artifact download URL still can only be obtained using this way, it's recommended to restore this functionality; unless links to artifacts can be made some other way.

.github/workflows/build.yml
```diff
@@ -144,116 +143,67 @@ jobs:
     - name: Build Cmder Launcher
       shell: pwsh
       working-directory: scripts
-      run: .\build.ps1 -Compile -verbose
+      run: .\build.ps1 -Compile -verbose -Terminal all
```
DRSDavidSoft: Moved `-verbose` to the end of the line so the command order stays consistent with the rest of the workflow calls.

.github/workflows/build.yml
DRSDavidSoft: Restored the artifact download link in the package summary by using `GH_TOKEN` plus `Get-ArtifactDownloadUrl` again.

DRSDavidSoft: Restored the previous per-file download behavior in the summary: each artifact row now links to its own uploaded artifact again, and the standalone "Download the combined build artifact" sentence is gone.

I also kept the emoji headings and the backticked profile names, and used unique internal artifact names for each profile's `hashes.txt` upload so the links stay unambiguous.

launcher/src/strings.rc2
```diff
@@ -6,7 +6,7 @@ STRINGTABLE
 {
 	IDS_TITLE             "Cmder Launcher"
 
-	IDS_SWITCHES          L"Valid options:\n\n    /c [CMDER User Root Path]\n    /task [ConEmu Task Name]\n    /icon [CMDER Icon Path]\n    [/start [Start in Path] | [Start in Path]]\n    /single\n    /m\n    /x [ConEmu extra arguments]\n\nor, either:\n    /register [USER | ALL]\n    /unregister [USER | ALL]"
+	IDS_SWITCHES          L"Valid options:\n\n    /c [CMDER User Root Path]\n    /task [Windows Terminal Profile/ConEmu Task Name]\n    /icon [CMDER Icon Path] - ConEmu ONLY!\n    [/start [Start in Path] | [Start in Path]]\n    /single - ConEmu ONLY!\n    /m\n    -- [ConEmu/Windows Terminal extra arguments]\n\nNote: '-- [...]' must be the last argument!\n\nor, either:\n    /register [USER | ALL]\n    /unregister [USER | ALL]"
```
DRSDavidSoft, line 9: Need to adjust the help message to separate out the console-specific flags.

The current " - * ONLY!" message, while accurate, is less UX-friendly than could be, especially using the current TaskDialog design.

I need to work on it once `development` is merged into `master`.

scripts/build.ps1
```diff
@@ -56,11 +56,14 @@ Param(
     # Using this option will skip all downloads, if you only need to build launcher
     [switch]$noVendor,
 
+    # Using this option will specify the emulator to use [none, all, conemu-maximus5, or windows-terminal]
```
DRSDavidSoft, line 59: What I need to do next:

1. Use the phrase "terminal emulator" instead of "emulator"
2. Refer the user to the new `packaging-profiles.json` infrastructure

scripts/build.ps1
```diff
@@ -56,11 +56,14 @@ Param(
...
+    [string]$Terminal = 'all',
+
     # Build launcher if you have MSBuild tools installed
     [switch]$Compile
 )
 
-# Get the scripts and Cmder root dirs we are building in.
+# Get the scripts and cmder root dirs we are building in.
```
DRSDavidSoft: No need to make the application's name lower case here, "Cmder" was just fine, and should be used everywhere as branding is capitalized

scripts/build.ps1
```diff
@@ -113,15 +117,32 @@ if (-not $noVendor) {
...
+            $WinTermSettingsJsonSave = Join-Path $config "windows_terminal_settings.json"
+            Write-Verbose "Backup '$WinTermSettingsJson' to '$WinTermSettingsJsonSave'"
+            Copy-Item $WinTermSettingsJson $WinTermSettingsJsonSave
+        }
+        else { $WinTermSettingsJson = "" }
+    }
+    else { $WinTermSettingsJson = "" }
+
     # Kill ssh-agent.exe if it is running from the $env:cmder_root we are building
+    $cmder_folder = $cmder_root.toString()
```
DRSDavidSoft: Investigate whether this could still be a cast to string without requiring to declaring a new `$cmder_folder` variable, in addition to `$cmder_root`, making below a one-liner again without actually breaking the build script like it used to.

If not, I think a more suitable name should be used for it.

daxgames:
See my last comment in https://github.com/cmderdev/cmder/pull/3090#issuecomment-4703569067

I edited it like 20 times so just want to make sure you see it in its final glory  @DRSDavidSoft.

We should continue the discussion here since this pr is actually open.

daxgames:
@DRSDavidSoft The new build script fails to run on My Windows 11 machine.

Copilot said it was cosed by unicode emojis:

image

It runs now but I did not want to commit/push this.

DRSDavidSoft:
@daxgames Working on it right now with Codex; seems like PowerShell 5.1 doesn't like UTF-8 encoded scripts without BOM while PowerShell 7 allows it.  Since both my machine and GitHub actions' infrastructure ran on pwsh v7, adding the emojis didn't cause an issue there.  So I'm applying the BOM now, which would hopefully fix the problem and allow the CI/CD to keep the emojis.  Sorry for the regression! 😅 

Nice diff by the way, how was it generated?

Back to the previous discussion we were having soon 🙏🏻

DRSDavidSoft:
@daxgames thanks for catching this. I reproduced the Windows PowerShell 5.1 failure locally: the emojis themselves are valid, but `scripts/utils.ps1` was UTF-8 without BOM, so Windows PowerShell decoded it as the legacy ANSI codepage and the emoji bytes became mojibake that broke parsing.

I pushed a fix in 8155738 that keeps the emojis by making `scripts/utils.ps1` UTF-8 with BOM. While testing the build script locally, I also found/fixed a Windows PowerShell 5.1 issue where `scripts/build.ps1` default paths based on `$PSScriptRoot` evaluated too early under `powershell.exe -File`.

Could you please share:

- exactly how you invoked the build script, including shell, command, and working directory
- any logs/output from the failing run before your local workaround
- what generated the nice side-by-side diff in the screenshot you posted

> Comment posted by ChatGPT Codex on behalf of the user (@DRSDavidSoft)

daxgames:
I use `delta` as my git pager.  You can cind it on github. I also use `bat` as  replacemwnt for `cat` that looks very similar to `delta` in that it does colored syntax when displaying a file.  Both are pretty cool.

As far as the other asks.

I do not have the failed run output handy but if you need it I can get it.

I ran the command `./scripts/build.ps1 -verbose -compile`

Shell was Visual Studio 2022 Powershell x64 terminal.

DRSDavidSoft:
@daxgames Thanks dax, I hope the build script is now working correctly as before the regression.

Will switch to the ongoing previous discussion from https://github.com/cmderdev/cmder/pull/3090#issuecomment-4703569067 soon; as agreed we'll continue discussion from the old PR here after I fully read and review your excellent points.

Will notify soon 🙏🏻

vendor/user_init.cmd.template
```diff
@@ -0,0 +1,163 @@
+@echo off
+
+:: This file was autogenerated by Cmder init.bat
```
DRSDavidSoft: Codex, to make the intent more clear, let's add the word **initially** like this:

This file was initially autogenerated by Cmder init.bat

vendor/user_init.cmd.template
```diff
@@ -0,0 +1,163 @@
...
+::
+:: It is yours to edit and will not be touched again by Cmder.
+::
+:: If you wish to recreate this file simply rename it and Cmder will re-create it the next time it is run
+:: or run the followin command from a Cmder shell:
```
DRSDavidSoft: Codex, please apply typo corrections to "followin" and other possible ones where required

daxgames:
@DRSDavidSoft Questions:

1. Why are you still persisting on changing `init.bat` to `init.cmd` it is COMPLETELY unnecessary.
2. The smallest distro is still called `slim_mini`, I think I have expressed my feelings on this name, I even offered alternative names that actually make sense and describe what the package contains.

    - Cmder for Windows: cmder_for_windows.zip
    - Cmder Launcher: cmder_launcher.zip

I have looked at your latest merges to `development` you seem to be attempting to re-invent the whell by writing, rewriting, and rewriting again the same workflows.  Theses same workflows have been in a state of flux for many, many months.

They are NEVER going to be perfect.  You said a couple of weeks ago you wanted one more 1.x.x release, presumably `v1.4.0` but we are still at `v1.3.25` and the pursuit of perfection is still delaying a release.

Can you just get `master` to where you want it with documented deprecation, a pending v2.0.0 breaking change warning in the `README.md` and lets release `v1.4.0`?

Possible future actions to speed releases while maintaining 1.x.x code base:

1. Copy `master` to `main` as you have suggested.
    - `master` can be retained for a while if necessary for `v1.4.x` BUGFIXES ONLY.
    - `Remove the `branches` and maybe `vendor` workflows from `master`.
    - Possibly add a `_1x` or `_legacy` suffix to build packages in a parallel release of `v2.x.x` to distinguish them from current.
2. Modify `development` so its intended merge target is `main`.
    - Change all references from `master` to `main`.
3. Merge `development` into `main`.
4. Release `v2.x.x`.

I don't mean to sound like a broken record but I just want to release something.

daxgames:
This is why changing `init.bat` to `init.cmd` for NO REASON is a HORRIBLE IDEA!  All I did is sync the latest `development` branch and try to start a new shell.

{060919B9-8CDB-4E43-8862-DB84C157FC8D}

daxgames:
After a little research on the deep technical differences between `.bat` and `.cmd` the `init.bat -> init.cmd` change needs to be retained but modified for better user experience.

The ONLY way I can see this working is to include the shim in the repo rather than creating it on on first run with a nag message with a link to docs on how to update his/her terminal config via the gui to use `init.cmd` instead of init.bat`.  The nag message needs to be in `init.bat` so it is only ever shown if `init.bat` is called.

There should be no instruction to delete `init.bat` because it will be put back on EVERY future update to maintain backward compatibility.  When the user changes their terminal config the nag goes away, no need to delete anything.

daxgames:
> After a little research on the deep technical differences between `.bat` and `.cmd` the `init.bat -> init.cmd` change needs to be retained but modified for better user experience.
> 
> The ONLY way I can see this working is to include the shim in the repo rather than creating it on on first run with a nag message with a link to docs on how to update his/her terminal config via the gui to use `init.cmd` instead of init.bat`. The nag message needs to be in `init.bat`so it is only ever shown if`init.bat` is called.
> 
> There should be no instruction to delete `init.bat` because it will be put back on EVERY future update to maintain backward compatibility. When the user changes their terminal config the nag goes away, no need to delete anything.

I have made the changes to allow migration to `init.cmd` while using an included `init.bat` backward compatibility shim for a better user experience. See #3104

DRSDavidSoft:
@daxgames

I'm preparing the remaining work as discussed and a proper reply to the points you raised in a separate comment. Here are some quick notes while working on this to avoid keeping this left unanswered.
- Will remove the "slim" name in the commits I'm preparing to push and apply the other fixes you pointed out too and revert regressions that I might have caused.
- Appreciate the bat/cmd investigation - I agree with your points, will answer all soon, as I have some notes to add about this topic.
- Focusing on getting the new release done ASAP and later working on the "legacy" branch that takes the place of the current master instead as requested, to avoid blocking the merge and making a new release.
- Doing some syncs between edits on my local machine in the meanwhile as I have unfinished work here, pulling the latest changes.
- Will be back with a detailed reply to your points with the work done.

Also, with the new pipeline and the CI/CD tasks, it will be easy to release different branches at the same time, so it makes sense to release a 2.0 before one more 1.x

Since 2.0 will semantically be a breaking change, I'll limit the scope of further updates to the dev branch until this PR is done, these will land in 2.1, 2.2, etc.

> I don't mean to sound like a broken record but I just want to release something.

My apologizes, you're completely right and I appreciate it. The blame is on me as you correctly pointed out. Working to resolve the issues now, including reducing the perfectionism. 🙏🏻

PR #2942, opened 2024-05-30
Cmder for Windows  - Add `Bash`, `Powershell`, `mintty`.

daxgames:
### Adds

- When no external terminal emulator is available.
  - Add `Cmder.exe /task cmder`.
  - Add `Cmder.exe /task bash`.
  - Add `Cmder.exe /task mintty`.
  - Add `Cmder.exe /task powershell`.
  - To run as Admin just create a shortcut and `Right Click -> Run as Administrator`.

image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,6 +52,20 @@ jobs:
       working-directory: scripts
       run: .\pack.ps1 -verbose -terminal all
 
+    - name: Upload artifact (cmder_win_mini.zip)
+      uses: actions/upload-artifact@v3
+      with:
+        path: build/cmder_win_mini.zip
+        name: cmder_win_mini.zip
+        if-no-files-found: error
+
+    - name: Upload artifact (cmder_win.7z)
+      uses: actions/upload-artifact@v3
+      with:
+        path: build/cmder_win.7z
+        name: cmder_win.7z
+        if-no-files-found: error
+
     - name: Upload artifact (cmder_win.zip)
       uses: actions/upload-artifact@v3
       with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,6 +52,13 @@ jobs:
       working-directory: scripts
       run: .\pack.ps1 -verbose -terminal all
 
+    - name: Upload artifact (cmder_win.zip)
+      uses: actions/upload-artifact@v3
+      with:
+        path: build/cmder_win.zip
+        name: cmder_win.zip
+        if-no-files-found: error
+
     - name: Upload artifact (cmder_wt.zip)
       uses: actions/upload-artifact@v3
       with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -45,12 +45,12 @@ jobs:
     - name: Build Cmder Launcher
       shell: pwsh
       working-directory: scripts
-      run: .\build.ps1 -Compile -verbose
+      run: .\build.ps1 -Compile -verbose -terminal all
 
     - name: Pack the built files
       shell: pwsh
       working-directory: scripts
-      run: .\pack.ps1 -verbose -emulator all
+      run: .\pack.ps1 -verbose -terminal all
 
     - name: Upload artifact (cmder_wt.zip)
       uses: actions/upload-artifact@v3
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ on:
     tags:
       - "v*"
   pull_request:
-    branches: [ "master" ]
+    branches: [ "master", "development" ]
 
 #---------------------------------#
 #    environment configuration    #
@@ -40,7 +40,7 @@ jobs:
         fetch-depth: 0
 
     - name: Add MSBuild to PATH
-      uses: microsoft/setup-msbuild@v1
+      uses: microsoft/setup-msbuild@v2
 
     - name: Build Cmder Launcher
       shell: pwsh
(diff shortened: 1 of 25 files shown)


daxgames:
So this is now broken on my Windows 11.

`cmder.exe /task cmd` - Works great!
`cmder.exe /task powershell` - Works great!
`cmder.exe /task bash` - Works NOT!
`cmder.exe /task mintty` - Works NOT!

No idea why.

Double clicking on the below works:

image

And these are what cmder.exe uses to launch `bash` and `mintty`.

Need to run them differently I guess but I don't know how.  I believe it works on Windows 10 not certain though.

DRSDavidSoft:
@daxgames It appears you're essentially opening `mintty.exe` which I believe is a terminal emulator in of itself (like `ConEmu.exe` and `conhost.exe` (the "Windows console") already are) which doesn't make sense.

We open the shell (e.g. `cmd.exe` or `bash.exe`) in those terminals, e.g. mintty + bash. So the solution would be NOT to do this in your code:
```
"%ProgramFiles%\git\usr\bin\mintty.exe" /bin/bash -l
```

And do this instead:
```
"%ProgramFiles%\git\usr\bin\bash.exe" -l
```

I don't believe it would make a difference on Win10/Win11 though.

I'm a bit tight on time so I can't exactly verify, but at my workplace whenever I need to use bash, I always call `bash.exe` from the PATH, and if I call `mintty.exe` instead, I'll always get the [mintty](https://mintty.github.io/) terminal emulator instead.

Sidenote, if you'd like some Windows 10 VMs, let me know so I can set one up for the you so you can clone the development version of Cmder in it for testing, if it helps. I can spin one up at work for this purpose.

daxgames:
@DRSDavidSoft the last time I tested this on Windows 10 it all worked perfectly.  I would not have opened the PR if it hadn't.

One thing I noticed on Windows 11 testing that I hadn't seen on Windows 10 is that cmd and powershell are opening up in Windows Terminal instead of cmd.exe and powershell.exe 'legacy' windows.

daxgames:
Here it is working on Windows 10 today same code:

image

daxgames:
I think the difference is Windows 11 prefers Windows Terminal and Windows 10 prefers conhost.exe????

daxgames:
> @daxgames It appears you're essentially opening `mintty.exe` which I believe is a terminal emulator in of itself (like `ConEmu.exe` and `conhost.exe` (the "Windows console") already are) which doesn't make sense.

You are missing the original point of Cmder for Windows, aka Cmder Micro, was to launch the terminal emulators that are already there.  It just does not work on Windows 11 as it does on Windows 10.

DRSDavidSoft:
@daxgames I believe the idealistic approach is for Cmder's launcher to hand over the task of choosing the terminal emulator (be it WT, `conhost`, ConEmu, `mintty`) to the OS, which currently is not case. It feels to me that we should either specifically run a **shell** or run a **shell** + **terminal** from the launcher, not do one case for cmd/powershell and the other for bash.

So, preferably, we should off hand the **terminal** part to OS and just launch the **shell**. The code needs to be modified for this so that `Cmder.exe` will launch a shell in the current terminal, or spawn a new terminal if it's launched from the GUI.

Alternatively, we should launch a terminal and run the shell in it. This can be `mintty`, `conhost`, WT, ConEmu.

In my humble opinion it doesn't make sense to bind bash to mintty. Sure, git-for-windows and mingw install both, but it doesn't mean that bash is exlusive to mintty. Bash can be ran in WT & conhost, and mintty can host clink, powershell, etc.

This is just my opinion, I'll be glad to hear what you think.

daxgames:
I like the fact there is an embedded terminal emulator and there always should be.  Cmder is after all a portable shell env.  I also like that there are currently choices of embedded terminal emulators.  I don't believe it should just be up to whatever is installed and available on the OS unless that is what the user wants.

I think your previous reply is what I said in my previous post but to be clear this PR attemps to:

In the absence of a `Cmder` embedded terminal emulator, launch the selected shell in whatever emulator is currently available whilt maintaining the ability to launch the same session types as prior releases of Cmder[cmd, powershell, bash, mintty(bash)].

> So, preferably, we should off hand the **terminal** part to OS and just launch the **shell**. The code needs to be modified for this so that `Cmder.exe` will launch a shell in the current terminal, or spawn a new terminal if it's launched from the GUI.
>
> Alternatively, we should launch a terminal and run the shell in it. This can be `mintty`, `conhost`, WT, ConEmu.

Basically above is what happens in Windows 10.

The code in this PR does the following if there is no `Cmder` embedded terminal emulator:

1. Launching `cmder /task powershell` runs `PowerShell.exe -ExecutionPolicy Bypass -NoLogo -NoProfile -NoExit -Command "Invoke-Expression 'Import-Module ''%CMDER_ROOT%\vendor\profile.ps1'''"`
2. Launching `cmder /task cmd` runs `cmd.exe -k "%cmder_root%\vendor\init.bat"`
3. Launching `cmder /task bash` runs `cmd.exe -c "%CMDER_ROOT%\vendor\start_git_bash.cmd"`
    * `-k` on Windows 10 left an orphaned `cmd.exe` window.
    * This is the same script I have proposed we use to launch bash from ConEmu or WT because it allows Cmder Mini to launch bash from well known locations.
    * A better approach may be to have the launcher do the things the `start_git_bash.cmd` script is doing.
4. Launching `cmder /task mintty` runs `cmd.exe -c "%CMDER_ROOT%\vendor\start_git_mintty.cmd"`
    * `-k` on Windows 10 left an orphaned `cmd.exe` window.
    * Similar to `%CMDER_ROOT%\vendor\start_git_bash.cmd`, finds `mintty` in well known locations and runs it.
    * Simply here for feature parity with previous `Cmder` releases.
    * A better approach may be to have the launcher do the things the `start_git_mintty.cmd` script is doing or as you hav suggested get rid of `mintty` completely.  More on this below.

> In my humble opinion it doesn't make sense to bind bash to mintty. Sure, git-for-windows and mingw install both, but it doesn't mean that bash is exlusive to mintty. Bash can be ran in WT & conhost, and mintty can host clink, powershell, etc.

I REALLY don't care about `mintty`, I don't use it, and no one said we should bind `bash` to it.  I was just trying to allow what `Cmder` does today minus the embedded terminal emulator and today `Cmder` can launch `bash` using `mintty` or simply 
 a `bash` session.  I did not want someone opening a bug report with the subject of 'Where's my `mintty` session in `Cmder Micro`?'

Anything you mentioned above and not currently in this PR is way beyond my knowledge of C++.  It is simply not worth the time it would take me to figure out something I will never use.  What I initially did works great on Windows 10 and I do not have the time required to figure out how to get a similar outcome on Windows 11+.

DRSDavidSoft:
Great poins, as always, @daxgames.

I need to investigate the difference between Windows 10 and Windows 11 regarding this, as the main support for 10 is basically ending in about a year or so according to Microsoft and it will enter extended support. Hopefully I'll understand the issue and solve it so it will work the same way it did on 10.

Regarding handing some of the batch tasks to the launcher, I agree with your points. Also, I have been wanting to do a re-write of the launcher with the intent of modularizing its code into re-usable sections for a while now. I would like to put my focus into the development branch and move some of the batch tasks into the launcher, if that helps. We can decide which functionality is best suited as a script and which one is best to be written in C++.

Thanks as always for implementing this! 👍🏻

daxgames:
@DRSDavidSoft @cmderdev/core I want to merge this.  This PR or at least a version of this PR has been open forever.  I recently made some changes and have tested it. 

The current PR seems to work on both Windows 10 and Windows 11.

Unless there are voiced objections, I will merge this on 1/12/2025 unless someone else wants to test it and merge it themselves.

PR #2987, opened 2024-12-31
Update LICENSE, fix copyright license year

JasonnnW3000:
Yearly copyright bump. Padding my changed LOC! 😛


daxgames:
Isn't copyright date supposed to be the first year a product was first published?  If so even 2017 is wrong and 2025 is even more wrong.

PR #2995, opened 2025-03-09
Fix #2980

### LINKED ISSUE #2980: [Bug] First time Git Installation error is misleading: The batch file cannot be found
### Description of the issue

On the initial setup of **Cmder Full** edition, when **git-for-windows** is performing its first time setup, the following error is displayed:

```
The batch file cannot be found.
```

This is safe and must be discarded, however, since it's displayed to the user, it sometimes lead to confusion and the user might get mislead.

### How to reproduce

1. I downloaded 
https://github.com/cmderdev/cmder/releases/download/v1.3.25/cmder.zip

2. started it, in post installation:

````
'C:\WINDOWS\system32\drivers\etc\hosts' -> '/etc/hosts'
'C:\WINDOWS\system32\drivers\etc\protocol' -> '/etc/protocols'
'C:\WINDOWS\system32\drivers\etc\services' -> '/etc/services'
'C:\WINDOWS\system32\drivers\etc\networks' -> '/etc/networks'
The batch file cannot be found.
Creating user startup file: "C:\apps\cmder2\config\user_profile.cmd"
        1 file(s) copied.
```

daxgames:
Fix #2980

### DIFF
--- a/vendor/init.bat
+++ b/vendor/init.bat
@@ -443,7 +443,7 @@ if "%CMDER_CONFIGURED%" gtr "1" goto :CMDER_CONFIGURED
 if exist "%GIT_INSTALL_ROOT%\post-install.bat" (
     echo Running Git for Windows one time Post Install....
     pushd "%GIT_INSTALL_ROOT%\"
-    "%GIT_INSTALL_ROOT%\git-cmd.exe" --no-needs-console --no-cd --command=post-install.bat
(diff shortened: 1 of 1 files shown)


daxgames:
Tested LGTM
image

No `nul` file:

image

DRSDavidSoft:
Potentially this will hide all errors from the user as a side effect.

PR #2996, opened 2025-03-09
Update sources.json

daxgames:
- Update Git and Clink

### DIFF
--- a/vendor/sources.json
+++ b/vendor/sources.json
@@ -1,13 +1,13 @@
 [
   {
     "name": "git-for-windows",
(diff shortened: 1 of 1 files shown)


DRSDavidSoft:
I was surprised to see this, I wondered if my automatic updates (#2974) weren't working.

You can merge these automatic PRs that I have created too, it updates Clink to the same version.

Although, #2974 will update Git to a release candidate version, which might not be desirable and thus we can merge this one instead of that. If so, I might need to update the update script to ignore RC releases when opening a new PR.

Issue #3020, opened 2025-09-04
[Feature request] the hg prompt in cmd is very slow, it doesn't have async prompt support yet


chrisant996:
### Suggestion

Cmder's git prompt uses async prompt filtering in cmd (Clink) to make the prompt show up instantly.

Cmder's hg prompt doesn't use async prompt filtering yet.
Cmder's svn prompt doesn't use async prompt filtering unless a special config variable is set.

This was reported in chrisant996/clink#792 by @luc0x61.

### Use case

When the current directory is in a Mercurial or Subversion workspace, then the prompt is slow.

### Checklist

- [x] I have read the documentation and made sure this feature doesn't already exist.
- [x] I have searched for similar issues and found none that describe my feature request.

chrisant996:
I have a fix for these issues, and some other minor bugs in the hg and svn prompts as well.

I'll publish a PR soon.

chrisant996:
I installed both Mercurial and Subversion and tested the PR.

I found one additional pre-existing bug, and I updated this PR with another commit to fix it.  When the hg program is not found, then the PR fixed the Lua script error bug, but another bug caused ` ()` to be appended to the prompt.  It shouldn't append anything in that case, just like the svn prompt doesn't append anything in that case.

PR #3021, opened 2025-09-04
Fix cmderdev/cmder#3020; hg prompt is slow.

chrisant996:
Cmder's hg prompt didn't use async prompt filtering yet.
Cmder's svn prompt only used async prompt filtering if a special config variable was set (the commit which contributed it seems to have misunderstood the git config settings for the git async prompt).

This commit makes the following changes:
1.  Adds async prompt filtering for hg.
2.  Makes async prompt filtering for svn the default.
3.  Removes the prompt_overrideSvnStatusOptIn variable.
4.  Fixes a bug where any errors during `svn status` in the svn prompt accidentally show up in the terminal.
5.  Fixes a bug where any errors during `hg branch` in the hg prompt turn into Lua errors.
6.  Simplifies the code for colors in the hg and svn prompts.
7.  Clean up the svn prompt code and make it consistent with the git and hg prompt code.


chrisant996:
I tested these changes locally in a git dir, a hg dir, a svn dir, and a normal dir (no source control repo).

PR #3056, opened 2025-12-07
Update the Lua git prompt to support repos with git reftables.

chrisant996:
**This compensates for a recent 🚨 breaking change 🚨 in git:**

Git has changed how it stores refs, and how it stores the current branch info.  In a repo with reftables, Cmder shows the branch name as ".invalid".

The changes here update Cmder's Lua prompt (i.e. for Clink) to replace ".invalid" with "Loading..." until the git status command finishes in the background, and then it can update the prompt with the accurate branch name.

The changes here also move the " -> remote" logic from clink-completions/git_prompt.lua directly into the Cmder prompt implementation (otherwise it can't work properly with reftables).

There were also changes necessary in clink-completions to support git reftables, and the clink-completions repo has already been updated accordingly (and the git_prompt.lua script has been completely deleted since it really always belonged as built into Cmder anyway).

To test git reftables, you can use the latest version of git to create a new repo:  `git init test_repo && cd test_repo`.  Without these changes, the Cmder prompt will be incorrect.  With these changes, the Cmder prompt should briefly showing "Loading..." and then show the correct branch name.  It won't say "Loading..." again until you change to a different repo.

Or you can simulate the behavior by running `set CLINK_DEBUG_GIT_REFTABLE=1`.

### DIFF
--- a/vendor/clink.lua
+++ b/vendor/clink.lua
@@ -393,8 +393,13 @@ local function get_git_remote(git_dir, branch)
     local file = io.open(git_dir.."/config", 'r')
     if not file then return nil end
 
-    local git_config = {};
-    local section;
+    local git_config = {}
+
+    local function get_git_config_value(section, param)
+        return git_config[section] and git_config[section][param] or nil
+    end
+
+    local section
     for line in file:lines() do
         if (line:sub(1,1) == "[" and line:sub(-1) == "]") then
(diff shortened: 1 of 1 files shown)


chrisant996:
@daxgames @DRSDavidSoft bump.

PR #3060, opened 2025-12-14
Install pacman script

daxgames:
Optional ability to install `pacman` in Git-for-Windows.  I say optional because I do not believe `pacman` should be included in Cmder releases.

1. Add `scripts/install_pacman.sh` script to enable installation of Pacman package manager into Cmder.
    - Currently depends on [pacman-for-git](https://github.com/daxgames/pacman-for-git), forked from [pacman-for-git](https://github.com/mcgitty/pacman-for-git) so I had control of updating git version commit hashes.
        - We may want to fork this here, so we all have control of it if/when we include this in Cmder.
        - We will need to update this repo before updating the Cmder git version in `vendor\sources.json`.
      
2. Update `scripts/build.ps1` to have a `-InstallPacman` switch.
    - Did this temporarily for testing.
    - We could remove the switch and make it the default if we decide to make `pacman` a default part of `Cmder - Full`

    or

    - We could leave `Cmder - Full` as is and create a `Cmder - Ex`(Ex = Extendable) or `Cmder - Max` package.
      - Not recommending this but it is an option.
  
3. Update `vendor/sources.json`.
4. Add vscode workspaces files to `.gitignore`.


DRSDavidSoft:
@daxgames Haven't reviewed this in full yet, but this is an excellent and missing piece of functionality for Cmder and I'm super excited to integrate it as part of the Cmder (the same way PsGet and other PowerShell-related integrations worked)

Having to manually install MSYS2 on top of Git for Windows is annoying, so if this works reliably, my vote is to keep it as a first-class feature of Cmder.

Nice work @daxgames, and thank you. Will review and merge soon 👍🏻 ❤️

scripts/install_pacman.sh
```diff
@@ -25,8 +24,8 @@ if [[ "$HOSTTYPE" == "i686" ]]; then
     msys2-keyring-1~20210213-2-any.pkg.tar.zst
   )
 
- zstd=zstd-1.5.0-1-i686.pkg.tar.xz
- zstd_win=https://github.com/facebook/zstd/releases/download/v1.5.5/zstd-v1.5.5-win32.zip
+  zstd=zstd-1.5.0-1-i686.pkg.tar.xz
```
DRSDavidSoft, line 27: @daxgames

Added invalid indentation here accidentally, please remove

To avoid keeping track of a random version such as v1.5.5, I'm considering to apply some minor improvements

DRSDavidSoft:
@daxgames Excited to merge this soon, but before that I'm going to address the current dev branch PR, soon

Thanks 🙏🏻

PR #3088, opened 2026-06-14
Invalid

daxgames:


DRSDavidSoft:
@daxgames Note that one of my comments about the pack.ps1, is the repeated build flags for each flavor of Cmder; I'll later push a commit that would avoid duplicate code and keep it [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

**Update:** landed in https://github.com/cmderdev/cmder/pull/3090

PR #3018, opened 2025-08-20
Added Clink ARM support

tlm2021:
Added support for ARM architecture in `vendor/init.bat`

PR #3096, opened 2026-06-21
revert unnecessary changes.

daxgames:


DRSDavidSoft:
@daxgames Thank you for the clean PR to fix my mistake 🙏🏻 

The `%print_debug%` removals are especially a large chunk of the bloat, I'm just skimming through any other additions I've caused then merge this before continuing on `development`

vendor/user_init.cmd.template
```diff
@@ -25,7 +29,6 @@ goto :SKIP_CLINK
 
   :: Check if Clink is not present
   if not exist "%CMDER_ROOT%\vendor\clink\clink_%clink_architecture%.exe" (
-    %print_error% "Clink executable is not present in 'vendor\clink\clink_%clink_architecture%.exe'"
```
DRSDavidSoft, line 28: @daxgames Assuming the Cmder directory content hasn't changed, correct?

Also, I'm considering whether the `if` branch in batch causes slow downs, if it does, I need to pay more attention to usages like this in other places as well

vendor/user_init.cmd.template
```diff
@@ -41,24 +44,14 @@ goto :SKIP_CLINK
...
-  :: Cleanup legacy Clink Settings file
-  if exist "%CMDER_CONFIG_DIR%\settings" if exist "%CMDER_CONFIG_DIR%\clink_settings" (
-    del "%CMDER_CONFIG_DIR%\settings"
-  )
-
-  :: Cleanup legacy Clink history file
-  if exist "%CMDER_CONFIG_DIR%\.history" if exist "%CMDER_CONFIG_DIR%\clink_history" (
-    del "%CMDER_CONFIG_DIR%\.history"
-  )
-
```
DRSDavidSoft, line 53: @daxgames These legacy clean ups are one time jobs that meant to be executed from `init.bat` once instead of every time by `init_user.cmd`, correct?

vendor/user_init.cmd.template
```diff
@@ -70,85 +63,66 @@ goto :SKIP_CLINK
...
   prompt `$E[1;32;49m`$P`$S`$_`$E[1;30;49mλ`$S`$E[0m
 
-  :: Add Windows Terminal shell integration support (OSC 133 sequences)
-  if defined WT_SESSION (prompt `$e]133;D`$e\`$e]133;A`$e\`$e]9;9;`$P`$e\%PROMPT%`$e]133;B`$e\)
-
   chcp %cp%>nul
 
 :CLINK_FINISH
+  :: Add Windows Terminal shell integration support (OSC 133 sequences)
+  if defined WT_SESSION (prompt `$e]133;D`$e\`$e]133;A`$e\`$e]9;9;`$P`$e\%PROMPT%`$e]133;B`$e\)
```
DRSDavidSoft, line 58: @daxgames I think these OSC definitions should be moved before the `:CLINK_FINISH` and still be kept under `:SKIP_CLINK`, because Clink shells don't make use of native Windows `prompt` anyway, so they might add to the unnecessary command execution

vendor/user_init.cmd.template
daxgames: yes

vendor/user_init.cmd.template
daxgames: ok - you know more about it than I do.

DRSDavidSoft:
Will come back to this PR later, have to take care of some tasks 🙏🏻 
Appreciate the help with aligning the init scripts; I should have worked on these much sooner

vendor/user_init.cmd.template
DRSDavidSoft: Rule of thumb, `clink` never uses the native Windows prompt.

In fact the `prompt <...>` command was removed in Cmder circa introduction of Clink, which is excellent.  However on platforms where clink wasn't working (e.g. ARM) the native `prompt` was added as a fallback to avoid making the shell look vanilla instead of our own custom prompt.

It is still useful in my opinion when Clink isn't available, or disabled due to users' preferences.

daxgames:
still cleaning it up - do not merge yet

vendor/user_init.cmd.template
```diff
@@ -70,85 +63,66 @@ goto :SKIP_CLINK
...
+  )
 
-%print_debug% init.cmd "START - nix_tools(%path_position%): Env Var - PATH=%path%"
-set "git_mingw_bin="
-if exist "%GIT_INSTALL_ROOT%\mingw32" (
-  set "git_mingw_bin=%GIT_INSTALL_ROOT%\mingw32\bin"
-) else if exist "%GIT_INSTALL_ROOT%\mingw64" (
-  set "git_mingw_bin=%GIT_INSTALL_ROOT%\mingw64\bin"
-)
+  :: TODO: Support for ARM
```
DRSDavidSoft, line 83: Related ARM for Clink are in https://github.com/cmderdev/cmder/pull/3018, I'm looking forward to getting that merged into `development` as well, especially with the relevance of more and more ARM machines, like Microsoft's own Surface line

Keeping these here for my own reference
- https://www.msys2.org/docs/arm64/
- https://github.com/git-for-windows/git/releases/ (arm releases)

PR #3100, opened 2026-06-30
Fix vendor PR status and branch sync

DRSDavidSoft:
## Summary

- Replace the redundant per-vendor release-version shield with a separate `Build Status` table column.
- Use compact Shields.io GitHub check-runs badges linked to each vendor repository's Actions page.
- Restore the top `Release notes · Compare changes` link row inside each vendor details block.
- Remove the misleading bottom blockquoted compare link, since it compared git commits rather than release-note ranges.
- Add explicit `actions: write` and `contents: write` permissions to the `Update branches` workflow so the master-to-development sync can push merge commits that include workflow-file changes.

## CI root cause

Recent post-merge `Update branches` runs failed while pushing `master` into `development`:

> refusing to allow a GitHub App to create or update workflow `.github/workflows/build.yml` without `workflows` permission

The merge itself is clean; the push is rejected because the merge includes `.github/workflows/*` changes.

## Validation

- `git diff --check`
- Executed the updated vendor PR-body generation logic locally against the #3093 version set and verified:
  - fourth `Build Status` column exists
  - four vendor-specific `<details>` blocks exist
  - four check-runs status shields exist
  - no GitHub release-version shields remain
  - `Release notes · Compare changes` is restored
  - no bottom `View full changes between...` commit-diff link remains
- Dry-ran `git merge --no-ff upstream/master --no-commit` into `upstream/development`; merge was conflict-free.

### DIFF
--- a/.github/workflows/branches.yml
+++ b/.github/workflows/branches.yml
@@ -9,6 +9,10 @@ on:
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
+permissions:
+  actions: write
+  contents: write
+
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
   # This job updates the development branch with the master branch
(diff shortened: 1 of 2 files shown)


walterdealmeida973-debug:
E AI

